PR #3, opened 2024-05-10
Initialize the repository structure

daisy-ycguo:
## Type of Change

Initialize the repository structure

## Description

Initialize the repository structure based on discussion


README.md
```diff
@@ -1,3 +1,14 @@
 # GenAIInfra
 
-Containerization and cloud native suite for OPEA
+GenAIInfra is the containerization and cloud native suite for OPEA, which includes artifacts to deploy OPEA in a cloud native way.
```
ftian1, line 3: Although there have had the word of "includes artifacts to deploy OPEA", shall we emphasize the main purpose of this repo is used by enterprise to deploy GenAIExamples to their own cloud? 

Maybe we need a flow/chart to demonstrate this relationships between different OPEA repos.

daisy-ycguo: > Although there have had the word of "includes artifacts to deploy OPEA", shall we emphasize the main purpose of this repo is used by enterprise to deploy GenAIExamples to their own cloud?
Good idea.
> 
> Maybe we need a flow/chart to demonstrate this relationships between different OPEA repos.
It's a good suggestion. I just wonder if we should place the relationship chart here or some place else, for example the doc repo. Because this repo is for deployment only and the doc repo is for the whole OPEA project. Can I create an issue at doc repo to track this question and use a separate PR for the relationship chart?

daisy-ycguo: @ftian1 I updated according to your first suggestion. Because this repo will support both k8s deployment and docker compose deployment. I think the target users of this repo are not only enterprise users but also personal users. So I try to cover both of these two kinds in the document. 

Can you review it again? Thank you.

ftian1: sure. we can use a separate PR to update the work flow. The change is ok for me

yongfengdu (approved):

PR #8, opened 2024-05-13
[WIP] Add helm lint check to github workflows

daisy-ycguo:
## Type of Change

Add a helm lint check job to github workflows

## Description

Add a helm lint check job to github workflows including helm install and helm lint check


jfding:
Please fix the "pre-commit.ci" checking failures

PR #10, opened 2024-05-14
Add helm lint check in github workflow

daisy-ycguo:
## Type of Change

Add helm lint check in github workflow

## Description

Add helm lint check in github workflow

### DIFF
--- a/.github/workflows/chart-validate.yaml
+++ b/.github/workflows/chart-validate.yaml
@@ -1,0 +1,54 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+name: Charts Validate
+
+on:
+  pull_request:
+    branches: [main]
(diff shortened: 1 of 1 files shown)


.github/workflows/chart-validate.yaml
```diff
@@ -0,0 +1,53 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+name: Charts Validate
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, reopened, ready_for_review, synchronize] # added `ready_for_review` since draft is skipped
+    paths:
+      - helm-charts/**
```
chensuyue, line 11: please add this line, so the test will be triggered in this PR. 
- .github/workflows/chart-validate.yaml

daisy-ycguo: I've added to the latest version. Please review again. Thank you.

chensuyue (approved):

Issue #76, opened 2024-06-05
Codegen service port is changed in GenAIExamples


daisy-ycguo:
CodeGen containerPort is changed from 6666 to 7778 in GenAIExamples: https://github.com/opea-project/GenAIExamples/pull/249
We should sync with them.
We should also check the CodeGen e2e test to make sure it works well.

PR #78, opened 2024-06-05
Update codegen service port

daisy-ycguo:
## Description

Update codegen service port and update e2e test

## Issues

fixes issue #76 

## Type of change

## Dependencies

n/a

## Tests

n/a


helm-charts/codegen/gaudi-values.yaml
```diff
@@ -13,10 +13,10 @@ image:
   # Overrides the image tag whose default is the chart appVersion.
   # tag: "1.0"
 
-port: 6666
+port: 7778
```
yongfengdu, line 16: According to the original change, it defined an environment var for customized port number and set 7778 as default:
MEGA_SERVICE_PORT = os.getenv("MEGA_SERVICE_PORT", 7778)

Although it's less possible user would define his own port value, it's better to support this choice to pass the "MEGA_SERVICE_PORT" to container in this template.
codegen/templates/deployment.yaml

yongfengdu (changes requested):

lianhao:
I would suggest split this changes into 2 different PR. One is for CI actions, it should succeed with any port settings in the helm-chart and/or manifest. Another is for populating MEGA_SERVICE_PORT as environment into k8s deployment defintion.

yongfengdu:
> I would suggest split this changes into 2 different PR. One is for CI actions, it should succeed with any port settings in the helm-chart and/or manifest. Another is for populating MEGA_SERVICE_PORT as environment into k8s deployment defintion.

The MEGA_SERVICE_PORT should not be exposed to users. User should care only the service.port, whose default is equal to MEGA_SERVICE_PORT but can be customized.
Let's first merge this fix to make sure we're using the correct port, and do more cleanups in another PR.

helm-charts/codegen/gaudi-values.yaml
daisy-ycguo: since I meet with error when adding this env, I will suggest we temporally comment the environment setting in this PR.

daisy-ycguo:
> I would suggest split this changes into 2 different PR. One is for CI actions, it should succeed with any port settings in the helm-chart and/or manifest. Another is for populating MEGA_SERVICE_PORT as environment into k8s deployment defintion.



> I would suggest split this changes into 2 different PR. One is for CI actions, it should succeed with any port settings in the helm-chart and/or manifest. Another is for populating MEGA_SERVICE_PORT as environment into k8s deployment defintion.

Will do.

helm-charts/codegen/gaudi-values.yaml
daisy-ycguo: Issue has been submitted to GenAIExamples: https://github.com/opea-project/GenAIExamples/issues/261

yongfengdu (approved):

lianhao (approved):
Let's merge this first and wait for the issue to be closed https://github.com/opea-project/GenAIExamples/issues/261, then creating another PR to cleanup them

PR #85, opened 2024-06-06
Enable golang e2e test in CI

daisy-ycguo:
## Description

Enable golang e2e test in CI

## Issues

n/a

## Type of change

## Dependencies

n/a

## Tests

CI test

### DIFF
--- a/.github/workflows/go-e2e.yaml
+++ b/.github/workflows/go-e2e.yaml
@@ -1,0 +1,71 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+name: Golang E2e Tests
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, reopened, ready_for_review, synchronize] # added `ready_for_review` since draft is skipped
+    paths:
+      - microservices-connector/**
+  workflow_dispatch:
+
+# If there is a new commit, the previous jobs will be canceled
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  GOSRC_DIR: "microservices-connector"
+
+jobs:
+  go-e2e:
+    runs-on: k8s
+    steps:
+      - name: Checkout out Repo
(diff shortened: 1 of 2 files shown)


lianhao (approved):

daisy-ycguo:
@irisdingbj If you want to add e2e tests for microservices-connector, please follow my guidance in scripts/e2e/go-test.sh and add your logic in three functions `install_gmc`, `validate_gmc` and `cleanup_gmc`. I've tested that any changes under the folder `microservices-connector` will trigger the go-e2e workflow. Please help to review and merge this PR. Then you can continue your development base on it. Thank you.

.github/workflows/go-e2e.yaml
```diff
@@ -0,0 +1,71 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  GOSRC_DIR: "microservices-connector"
+
+jobs:
+  go-e2e:
+    runs-on: k8s
```
irisdingbj, line 24: Thanks for the support for gmc e2e.   Since we need support Gaudi as well, possible to make this runner setting as an Env Var? then in the gmc test script we can leverage this env var to select and deploy related manifests.

.github/workflows/go-e2e.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Build image and push
+        run: |
+          cd $GOSRC_DIR
+          make docker.build
+          make docker.push
+
+      - name: Install modules
+        run: |
+          echo "should_cleanup=true" >> $GITHUB_ENV
+          .github/workflows/scripts/e2e/go_test.sh install_gmc
```
irisdingbj, line 49: Can we use separate script for different go project? In this case, CI can be clear which part fails.  Based on that, I will prefer naming it as `gmc_test.sh`. Thoughts?

irisdingbj:
Thanks for the support!  Let's merge this first and we can address the env var and filename in follow up PR.

irisdingbj (approved):

PR #140, opened 2024-06-28
Add dataprep common service

yongfengdu:
Add dataprep component

## Description

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/chatqna/Chart.yaml
+++ b/helm-charts/chatqna/Chart.yaml
@@ -30,5 +30,8 @@ dependencies:
   - name: retriever-usvc
     version: "0.1.0"
     repository: "file://../common/retriever-usvc"
+  - name: data-prep
+    version: "0.1.0"
+    repository: "file://../common/data-prep"
 version: 0.1.0
 appVersion: "1.0.0"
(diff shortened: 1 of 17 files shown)


daisy-ycguo (approved):
lgtm

helm-charts/chatqna/gaudi-values.yaml
```diff
@@ -9,25 +9,20 @@ replicaCount: 1
...
 
-securityContext:
-  allowPrivilegeEscalation: false
-  runAsNonRoot: true
-  runAsUser: 1000
-  capabilities:
-    drop:
-    - ALL
-  seccompProfile:
-    type: RuntimeDefault
```
lianhao, line 29: why remove default securityContext here?

helm-charts/chatqna/values.yaml
```diff
@@ -10,24 +10,19 @@ replicaCount: 1
...
-securityContext:
-  allowPrivilegeEscalation: false
-  runAsNonRoot: true
-  runAsUser: 1000
-  capabilities:
-    drop:
-    - ALL
-  seccompProfile:
-    type: RuntimeDefault
-
```
lianhao, line 30: ditto?

lianhao (changes requested):
Why remove the default security context in value files?

helm-charts/chatqna/gaudi-values.yaml
yongfengdu: I was working with an old version and these values was overwritten(Include the volume name change).
Will restore them in next revision.

lianhao (approved):

Issue #148, opened 2024-07-02
GMC: failed to write mcGraphHandler response


gyohuangxin:
I setup the GMC controller and codegen pipline, then I want to test if it works using command below:
```
accessUrl=$(kubectl get gmc -n codegen -o jsonpath="{.items[?(@.metadata.name=='codegen')].status.accessUrl}")
export CLIENT_POD=$(kubectl get pod -n default -l app=client-test -o jsonpath={.items..metadata.name})
kubectl exec "$CLIENT_POD" -- curl $accessUrl  -X POST  -d '{"query": "def print_hello_world():"}' -H 'Content-Type: application/json'
```
But it fails with error message:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    37    0     0    0    37      0      0 --:--:--  0:01:51 --:--:--     0
curl: (52) Empty reply from server
command terminated with exit code 52
```

The log of the router-service-deployment-5b87857fb6-txbcw pod:
```
{"level":"info","ts":"2024-07-02T08:24:17Z","logger":"GMCGraphRouter","msg":"Starting execution of step","type":"serviceUrl","stepName":"Llm"}
{"level":"info","ts":"2024-07-02T08:24:17Z","logger":"GMCGraphRouter","msg":"Entering callService","url":"http://llm-service.codegen.svc.cluster.local:9000/v1/chat/completions"}
{"level":"info","ts":"2024-07-02T08:24:17Z","logger":"GMCGraphRouter","msg":"Print the http request headers","HTTP_Header":{"Accept":["*/*"],"Content-Length":["37"],"Content-Type":["application/json"],"User-Agent":["curl/7.74.0"]}}
{"level":"info","ts":"2024-07-02T08:26:09Z","logger":"GMCGraphRouter","msg":"elapsed time","step":"http://llm-service.codegen.svc.cluster.local:9000/v1/chat/completions","time":111.759226138}
[...]

PR #149, opened 2024-07-02
Fix bugs for router on handling response from pipeline microservices

zhlsunshine:
## Description

There will be timeout for router to handle response from pipeline microservices due to it need to wait for the whole response.
Using bufio to handle the response and make the new goroutine to call microservices in `mcGraphHandler`

## Issues

https://github.com/opea-project/GenAIInfra/issues/148

## Type of change

## Dependencies

`n/a`.

## Tests

Below is the testing snapshot:

image


microservices-connector/cmd/router/main.go
```diff
@@ -365,22 +367,54 @@ func routeStep(nodeName string, graph mcv1alpha3.GMConnector, input []byte, head
...
+				return
+			}
 		}
+	}()
+
+	select {
+	case <-ctx.Done():
+		log.Error(errors.New("failed to process request"), "request timed out")
+		http.Error(w, "request timed out", http.StatusGatewayTimeout)
+	case <-done:
```
gyohuangxin, line 417: Could you also add some logs to suggest that the process is successful?

zhlsunshine: okay, make sense.

gyohuangxin (approved):
I've verified it worked on my environment.

microservices-connector/cmd/router/main.go
```diff
@@ -365,22 +367,55 @@ func routeStep(nodeName string, graph mcv1alpha3.GMConnector, input []byte, head
...
-			log.Error(err, "failed to write mcGraphHandler response")
+
+		writer := bufio.NewWriter(w)
+		defer func() {
+			if err := writer.Flush(); err != nil {
+				log.Error(err, "error flushing writer when processing response")
+			}
+		}()
+
+		for start := 0; start < len(response); start += 1024 {
```
mkbhanda, line 401: what is this magic number 1024 and incrementing by it?

microservices-connector/cmd/router/main.go
```diff
@@ -365,22 +367,55 @@ func routeStep(nodeName string, graph mcv1alpha3.GMConnector, input []byte, head
...
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(statusCode)
+			if _, err := w.Write(prepareErrorResponse(err, "Failed to process request")); err != nil {
+				log.Error(err, "failed to write mcGraphHandler response")
+			}
+			return
 		}
-	} else {
 		if json.Valid(response) {
 			w.Header().Set("Content-Type", "application/json")
```
mkbhanda, line 390: if json invalid what do we set in content-type? blank/nothing? is that valid?

mkbhanda: There seem to be several types looking at https://stackoverflow.com/questions/23714383/what-are-all-the-possible-values-for-http-content-type-header  and we might encounter these when using genAI for image, audio etc construction. Perhaps add a note "Other types not validated at this time" type comment.

mkbhanda (commented):
Please check inline comments. Happy to approve as soon as I hear from you.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @mkbhanda, the router service will write the response message back, and the router need to wait for the whole response from microservices. And this enhancement is that router service will write back as soon as there are some feedback from microservices. So I use 1024 as the length of the message to send it out.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @mkbhanda, it's a great question, router service as an entry endpoint, both the requests and responses of microservices are based on json format, so it's oaky for now. And I have some tests that the image can also be represented as base64 encode data.
Of course, there would be other changes when some other examples imported. Thanks for raising this question!

microservices-connector/cmd/router/main.go
mkbhanda: Just curious why you choose 1024, 1K. What are the repercussions if you had chosen 256 or 512 or 1 MB? Would it make sense to make this a constant or parameter, when dealing with different kinds of media and communication protocols for inter-microservice communication (like DMA, shared memory, UDP, TCP). After each such unit you are checking for errors.

mkbhanda (approved):
Thank you for the answers!

mkbhanda (approved):
LGTM, Thank you for introducing a constant, ChunkSize.

Issue #159, opened 2024-07-05
Optimize GMC e2e test in CI and reduce the test duration


daisy-ycguo:
Since GMC e2e is moved to KIND environment. The testing always takes 18+ minutes to finish. We should figure out some possible optimizations to reduce the test duration in order to improve efficiency

daisy-ycguo:
The GMC e2e on xeon has been reduced to 12 minutes. The test log is here: https://github.com/opea-project/GenAIInfra/actions/runs/9834081498/job/27145358124

daisy-ycguo:
The GMC e2e on Xeon has been reduced to 10+ minutes. The test log is here: https://github.com/opea-project/GenAIInfra/actions/runs/9834247322/job/27145780072

daisy-ycguo:
Fixed in this PR: https://github.com/opea-project/GenAIInfra/pull/160

PR #160, opened 2024-07-05
Optimize GMC e2e scripts

daisy-ycguo:
## Description

Optimize GMC e2e scripts to remove duplicated codes
Optimize GMC e2e workflow to reduce the latency

## Issues

closes issue #159

## Type of change

## Dependencies

n/a

## Tests

CI tests


.github/workflows/go-e2e.yaml
```diff
@@ -9,6 +9,8 @@ on:
     types: [opened, reopened, ready_for_review, synchronize] # added `ready_for_review` since draft is skipped
     paths:
       - microservices-connector/**
+      - manifests/ChatQnA/**
+      - manifests/DocSum/xeon/docsum*
```
irisdingbj, line 13: shall we add the other two examples(codegen and codetrans) as well?

irisdingbj (commented):
Thanks @daisy-ycguo for taking care of GMC e2e test.   Besides GMC e2e, will will move all e2e tests in kind environments?  Any problems we keep using the original k8s env on the CI/CD machine?

.github/workflows/go-e2e.yaml
daisy-ycguo: I think we've already added the test cases of codegen and codetrans. Yet GMC doesn't have any dependencies to manifest/CodeGen and manifest/CodeTrans. GMC only uses manifest/ChatQnA and manifests/DocSum/*/docsum_llm.yaml. So I just include these two folders in the workflow.

daisy-ycguo:
> Thanks @daisy-ycguo for taking care of GMC e2e test. Besides GMC e2e,Will we move all e2e tests in kind environments? Any problems we keep using the original k8s env on the CI/CD machine?

We just moved GMC e2e to KIND environment because GMC controller has to be unique in K8s and it blocks parallel testing of workloads in GenAIExamples/GenAIInfra.

By using this workflow: [gmc-on-push.yaml](https://github.com/opea-project/GenAIInfra/blob/main/.github/workflows/gmc-on-push.yaml), GMC images are build and opea-system are installed and updated to K8s cluster when GMC source codes are changed.

irisdingbj (approved):

yongfengdu (approved):

Issue #147, opened 2024-07-02
Redesign the way how common microservice installed and talk


bjzhjing:
There are two issues observed while try to install each microservice under ./heml-chars/common:

- Microservice like retriever-usvc has dependency specified in Chart.yaml, but embedding is missed there.
- Even if I install microservices - redis-vector-db and embedding-usvc firstly, retriever-usvc can not talk to them, it will deploy the former microservices itself and then talk to them.

Based on above issues, I'm thinking if we could install the common microservices separately. That is to install each microservice in order, and then they can talk to each other, rather than deploy the same service again and again.

yongfengdu:
Thanks for reporting this issue.
Originally there are less components, so hard dependency is set to ease the deployment configurations for GenAIExamples.
But with more components added, and user may want more flexible configuration during deployment, so it's time to remove these hard dependencies.
The helm chart dependency should only be set in the E2E application examples like ChatQnA
Will clean up this.

PR #161, opened 2024-07-08
Rm comp dep

bjzhjing:
## Description

Remove fixed dependencies for common microservices to allow end user to flexibly configure their workload pipeline.

## Issues

n/a

## Type of change

Enhancement for existing helm-chart design.

## Dependencies

n/a

## Tests

Issue 'helm install chatqna chatqna' to install chatqna workload to verify.


lianhao (changes requested):
Please also adjust the helm value files in all corresponding charts, otherwise parent helm chart's values won't be able to passed into child chart correctly.

yongfengdu:
closes: #147

lianhao (approved):

helm-charts/chatqna/gaudi-values.yaml
```diff
@@ -31,16 +31,17 @@ securityContext:
 
 # To override values in subchart llm-uservice
 llm-uservice:
+  TGI_LLM_ENDPOINT: ""
```
lianhao, line 34: maybe we should not set any thing for llm-uservice here, if the default one works for us

helm-charts/chatqna/values.yaml
```diff
@@ -31,10 +31,11 @@ securityContext:
 
 # To override values in subchart llm-uservice
 llm-uservice:
+  TGI_LLM_ENDPOINT: ""
```
lianhao, line 34: ditto

lianhao (approved):

daisy-ycguo (approved):
lgtm

PR #162, opened 2024-07-10
Add MAX_INPUT_TOKENS to tgi

yongfengdu:
New tgi 2.0.0 version will set MAX_INPUT_TOKENS to 4095 by default. This will cause tgi error. First set to 1024 as default, may make it configurable by user later.

## Description

## Issues

‘n/a'

## Type of change


lianhao (approved):

daisy-ycguo (approved):
lgtm

PR #176, opened 2024-07-16
support multiple nodes of a graph, apply different endpoint env for different node

KfreeZ:
## Description

#166 

this PR also adopt the new manifest, when new manifest is merged, some unnecessary code will be removed for simplicity

mark it as draft

## Issues

## Type of change

## Dependencies

## Tests


microservices-connector/config/samples/chatQnA_xeon.yaml
```diff
@@ -23,6 +23,7 @@ spec:
           serviceName: embedding-svc
           config:
             endpoint: /v1/embeddings
+            TEI_EMBEDDING_ENDPOINT: TeiEmbedding
```
irisdingbj, line 26: This should use servcieName, in this case `tei-embedding-svc`.

microservices-connector/config/samples/chatQnA_xeon.yaml
```diff
@@ -33,6 +34,7 @@ spec:
           serviceName: retriever-svc
           config:
             endpoint: /v1/retrieval
+            REDIS_URL: VectorDB
```
irisdingbj, line 37: same here, use `redis-vector-db`, in this way we can support multiple endpoint for switch scenario

microservices-connector/config/samples/chatQnA_xeon.yaml
```diff
@@ -43,6 +45,7 @@ spec:
           serviceName: reranking-svc
           config:
             endpoint: /v1/reranking
+            TEI_RERANKING_ENDPOINT: TeiReranking
```
irisdingbj, line 48: the same

microservices-connector/config/samples/chatQnA_xeon.yaml
```diff
@@ -55,6 +58,7 @@ spec:
           serviceName: llm-svc
           config:
             endpoint: /v1/chat/completions
+            TGI_LLM_ENDPOINT: Tgi
```
irisdingbj, line 61: the same

microservices-connector/config/samples/chatQnA_xeon.yaml
KfreeZ: In the multiple endpoint case, is it better to use different step name, like llm-big, llm-small?

mkbhanda (approved):
LGTM

KfreeZ:
use #203 to commit code

PR #186, opened 2024-07-18
helm: Update codegen to latest

lianhao:
## Description

- Update codegen chart to latest tgi-gaudi version and model

- Add helm test to codegen

## Issues

n/a.

## Type of change

## Dependencies

n/a.

## Tests

Manually verified.


lianhao:
The reason why gaudi test is failing is that for gaudi, we change the default model from `ise-uiuc/Magicoder-S-DS-6.7B` to `meta-llama/CodeLlama-7b-hf` for gaudi case to be in alligenment with GenAIExamples(The `ise-uiuc/Magicoder-S-DS-6.7B` model on latest tgi-gaudi will give a wired empty answer). 

However, the Huggingface token CI is using doesn't have access to download the model  `meta-llama/CodeLlama-7b-hf`. We need that token owner to goto [the page ](https://huggingface.co/meta-llama/CodeLlama-7b-hf) to apply for access first

yongfengdu (approved):

leslieluyu (approved):

Issue #207, opened 2024-07-22
support multipart/form-data in router


irisdingbj:
router is not able to handle multipart/form-data:

 kubectl exec -it client-test-87d6c7d7b-wvxf5 -n chatqa -- curl http://router-service-data-prep.chatqa.svc.cluster.local:8080  -X POST  -H 'Content-Type: multipart/form-data' -F 'link_list=["https://istio.io", "https://opea.dev"]'
{"detail":"Must provide either a file or a string list."}

irisdingbj:
close this as fix has been merged

PR #246, opened 2024-07-31
Add multiple endpoints for GMC pipeline via gmcrouter

zhlsunshine:
## Description

Add multiple endpoints for GMC pipeline via gmcrouter.
1. Support new router verb `dataprep` for gmcrouter
2. Support multipart/form-data header for gmcrouter

## Issues

https://github.com/opea-project/GenAIInfra/issues/168
https://github.com/opea-project/GenAIInfra/issues/207
https://github.com/opea-project/GenAIComps/issues/366

## Type of change

## Dependencies

`n/a`.

## Tests
**Query before Intel Gaudi information is injected:**
image

**Injecting the Intel Gaudi information**
image

**Query after Intel Gaudi information is injected:**
image
image

**There is no any dataprep router is defined:**
image


microservices-connector/cmd/router/main.go
```diff
@@ -477,6 +479,109 @@ func mcGraphHandler(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func mcDataHandler(w http.ResponseWriter, r *http.Request) {
+	defaultNode := mcGraph.Spec.Nodes[defaultNodeName]
+	for i := range defaultNode.Steps {
+		step := &defaultNode.Steps[i]
+		if DataPrep == step.StepName {
```
irisdingbj, line 486: There should be only one dataprep service,   so we can break out the loop if we found it

microservices-connector/cmd/router/main.go
```diff
@@ -477,6 +479,109 @@ func mcGraphHandler(w http.ResponseWriter, req *http.Request) {
...
+						return
+					}
+					_, err = io.Copy(part, file)
+					if err != nil {
+						http.Error(w, "Failed to copy file", http.StatusInternalServerError)
+						return
+					}
+				}
+			}
+			err = writer.Close()
```
irisdingbj, line 534: add writer into defer() block as well

microservices-connector/cmd/router/main.go
mkbhanda: this will get us past the dataprep hump if we break out I like the full pass through in case there are other endpoints. Perhaps multiple different datapreps for different kinds of data - audio, video, images, text, web search, some agent. May be even different embedding services with their own vector database paths.

Also do steps have some description text, so that when someone queries for these endpoints they know what to use for what, like "query path", "data upload path" or something better.

mkbhanda (commented):
@irisdingbj  and @zhlsunshine  I think there could be more than two endpoints .. please see my comment. Two is a reasonable default, need to being some context!

microservices-connector/cmd/router/main.go
```diff
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime/multipart"
```
KfreeZ, line 21: this is reminding me we should go mod tidy in each release cycle, @irisdingbj ?

microservices-connector/cmd/router/main.go
```diff
@@ -477,6 +479,109 @@ func mcGraphHandler(w http.ResponseWriter, req *http.Request) {
...
+				log.Error(err, "failed to copy response body")
+			}
+		}
+	}
+}
+
+func initializeRoutes() *http.ServeMux {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", mcGraphHandler)
+	mux.HandleFunc("/dataprep", mcDataHandler)
```
KfreeZ, line 581: maybe add a helper here? To indicate the correct path if user input unexpected ones?

KfreeZ:
A question: how we expose the /dataprep in router service's status?

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj, I think we need to consider that there are multiple dataprep microservices in Pipeline, such as in switch mode. So I suggest that we can keep this logic, does it make sense?

zhlsunshine: Hi @mkbhanda, the data formats supported by dataprep are only text format, such as txt, htm, md, etc. And for audio, video, images format, especially for video, we may need more time for supporting them because we need to handle the media stream in router. So suggest that we can just support text format as well, does it make sense?

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj ,  sure, missing this. Done!

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @KfreeZ, will update the go.mod

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @KfreeZ, I suggest that add a README for it, does it make sense?

microservices-connector/cmd/router/main.go
zhlsunshine: And some additional information, I handle all file in binary format, so in theories, it can handle all these kind of file format, but we need to handle the stream properly.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @KfreeZ, this library is the native lib of golang, so I understand you just want to comment with Iris about this. I thin k we can talk it on meeting.

KfreeZ (approved):

microservices-connector/cmd/router/main.go
mkbhanda: > maybe add a helper here? To indicate the correct path if user input unexpected ones?

@KfreeZ how would we do this? Need to keep all valid paths in some array and if user provides something not in valid list return a usage message?
Do we have a parent class from which our router paths are defined, perhaps we can add this valid paths functionality there. An enhancement feature?

microservices-connector/cmd/router/main.go
```diff
@@ -477,6 +479,112 @@ func mcGraphHandler(w http.ResponseWriter, req *http.Request) {
...
```
mkbhanda, line 584: To @irisdingbj comment about expecting only one dataprep path and @zhlsunshine allowing/expecting more, is this logic a simplification of adding route only for one dataprep path. We have to be consistent in our logic, support only one now and later more perhaps, or support multiple everywhere.

mkbhanda (commented):
Some inline comments. Will watch for you response and then approve.

microservices-connector/cmd/router/main.go
irisdingbj: can we add a listService endpoint by default in router? in this way user can list all available endpoint throuth router. because even for dataprep, it will have 3 endpoints, like /dataprep /getDoc /deleteDoc etc.

microservices-connector/cmd/router/main.go
irisdingbj: Let's target support more dataprep, in this case, we need generate handler dynamically instead of the fixed /dataprep one.

microservices-connector/cmd/router/main.go
```diff
@@ -477,6 +479,112 @@ func mcGraphHandler(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func mcDataHandler(w http.ResponseWriter, r *http.Request) {
+	defaultNode := mcGraph.Spec.Nodes[defaultNodeName]
```
irisdingbj, line 483: If there is no dataprep in the GMC config, this handler should return info like " this endpoint is not available" to end user.

microservices-connector/cmd/router/main.go
zhlsunshine: Yeah, this should be. So I original support the extra request Header of `SERVICE_NAME`, but remove it because I want to make the same format with `dataprep` microservice

microservices-connector/cmd/router/main.go
KfreeZ: my thought is like other http routers:
	mux.HandleFunc("/", mcGraphHandler)
	mux.HandleFunc("/dataprep", mcDataHandler)
        mux.HandleFunc("/*other invalid path*", helper) => indicate the legal paths

KfreeZ: according to my search, in the default go router :
    // Catch-all route for handling invalid paths
    router.NoRoute(func(c *gin.Context) {
        c.String(404, "Page not found")
    })

if that works, my thought is maybe we can add a helper here to indicate the correct path to access

microservices-connector/cmd/router/main.go
zhlsunshine: Yeah, I added the error feedback when there is no dataprep is found.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj, as we discussed, I will raise an issue for dataprep to DCAI team, they should remove such verb. It's not the best practice for REST API design.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj, after validation, `writer.Close()` should be operated before the request is sent out. So we can not put it in defer() block, otherwise, the request can not contain the input information, such file name or link path, etc.

zhlsunshine: And I also add `handleMultipartError` to handle the error occurred before `writer.Close()`

lianhao:
For the curl (52) error, please wait for PR #249 to land in with the workaround to unblock the issue #258

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj, I opened an issue for dataprep REST API definition: https://github.com/opea-project/GenAIComps/issues/394

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @KfreeZ, for now, we can just a remind for for user. Like in PR description:
image

zhlsunshine:
Hi @mkbhanda and @irisdingbj, for multiple dataprep REST API definition, I have opened an issue: https://github.com/opea-project/GenAIComps/issues/394 to track it. By following the best practice, there should be only one REST API for the microservice: `/v1/dataprep`, and for all other VectorDB, they really do have only one REST API `/v1/dataprep`. So I think let's see and keep current implementation. And I also add the logic for checking and providing information if there is no `dataprep` router. Does it make sense?

zhlsunshine:
Hi @mkbhanda and @irisdingbj, one more thing is that I remove the request size limitation because there is no size limitation in `dataprep-redis` microserivce now.

irisdingbj:
Let's address listService and dynamic adding endpoints seperately.

irisdingbj (approved):

PR #269, opened 2024-08-06
re-org k8s manifests files for GMC and examples

jfding:
## Description

Re-org and clean up of k8s manifests in this repo, including:
- moved manifests/common to GMC/config/manifests, and updated related docs and code
- clean up deprecated k8s manifest for genai-examples, in the future we'd like track them in GenAIExamples repo

## Issues

n/a

## Type of change

n/a

## Dependencies

## Tests


helm-charts/update_manifests.sh
yongfengdu, line 1: Since you have changed the location of the output yaml files, should modify this VAR accordingly.
OUTPUTDIR=${CUR_DIR}/../microservices-connector/config/manifests

And the location of update_manifests.sh is changed, it's better to replace all "../helm-charts/" with "./"

yongfengdu (approved):
LGTM

lianhao (approved):

zhlsunshine (approved):
I notice that the unit test is pass as well, LGTM!

Issue #272, opened 2024-08-06
GMC Router doesn't apply the parameters requested from client side


zhlsunshine:
GMC Router doesn't apply the parameters requested from client side, the below as an example:

```shell
curl http://router-service.chatqa.svc.cluster.local:8080  -X POST  -d '{"text":"What is the revenue of Nike in 2023?", "parameters":{"max_new_tokens":17, "do_sample": true}}' -H 'Content-Type: application/json'
```

The parameters of `max_new_tokens` does not take effect.

Found by @leslieluyu

zhlsunshine:
The PR got merged. So close this issue.

PR #273, opened 2024-08-06
Fix router bugs on max_new_tokens and dataprep gaudi yaml file

zhlsunshine:
## Description

Fix router bugs on max_new_tokens and dataprep gaudi yaml file. And remove `^M` from dataprep xeon yaml file.
This max_new_tokens bug was found when doing performance test by @leslieluyu 

## Issues

https://github.com/opea-project/GenAIInfra/issues/272

## Type of change

## Dependencies

 `n/a`.

## Tests

**Set max_new_tokens=17**
image

**Set max_new_tokens=50**
image
image


microservices-connector/cmd/router/main.go
```diff
@@ -198,6 +199,40 @@ func executeStep(
 	return callService(step, serviceURL, input, headers)
 }
 
+func mergeRequests(initReq, respReq []byte) []byte {
```
irisdingbj, line 202: there should be a place to add some predefined parameter value like https://github.com/opea-project/GenAIComps/blob/main/comps/cores/proto/docarray.py#L92

microservices-connector/cmd/router/main.go
```diff
@@ -198,6 +199,40 @@ func executeStep(
...
+	var initReqData map[string]interface{}
+	var respReqData map[string]interface{}
+
+	if err := json.Unmarshal(initReq, &initReqData); err != nil {
```
irisdingbj, line 206: please do not unmarshal intiReq every time to save time.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj, this is defined as `parameters` for all example, I have defined it in const.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @irisdingbj, okay, it makes sense, done!

microservices-connector/cmd/router/main.go
```diff
@@ -198,6 +199,32 @@ func executeStep(
...
+	var respReqData map[string]interface{}
+
+	if _, exists := initReqData[Parameters]; exists {
+		if err := json.Unmarshal(respReq, &respReqData); err != nil {
+			log.Error(err, "Error unmarshaling respReqData:")
+			return nil
+		}
+		// Merge init request into respReq
+		for key, value := range initReqData[Parameters].(map[string]interface{}) {
+			/*if _, exists := respReqData[key]; !exists {
```
KfreeZ, line 212: remove this if not needed

microservices-connector/cmd/router/main.go
```diff
@@ -198,6 +199,32 @@ func executeStep(
 	return callService(step, serviceURL, input, headers)
 }
 
+func mergeRequests(respReq []byte, initReqData map[string]interface{}) []byte {
```
KfreeZ, line 202: Just curious, why the handleEnsemblePipeline does not need this mergeRequest?

KfreeZ (approved):

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @KfreeZ, just let's keep it there, because let's see if they need to keep the response instead overwriting by initial request.

microservices-connector/cmd/router/main.go
zhlsunshine: Hi @KfreeZ, based on current implementation, EnsemblePipeline have no response case, so just let it there.

irisdingbj (approved):

Issue #128, opened 2024-06-25
Support production level volume types for model data


lianhao:
Currently, our k8s installation (either helm-charts or manifest) have hostPath volume type to store the model data to save model download time. This is NOT ready for production environment, and is NOT suitable for scale outing with multiple nodes.

In production environment, users often use persistent volume(PV/PVC) to store those kind of data. We should give users the options to choose which volume type they want to use.

For PVs, it should have annotation "pv.beta.kubernetes.io/gid: <actual gid>" so non root user in pod can have write access to the volume mounted from this PV. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#access-control as reference.

endomorphosis:
I agree with this, and this is one of the reasons why previous to OPEA I built huggingface model manager, that pulls models from ipfs / s3 / huggingface, and evicts models as needed.

daisy-ycguo:
@lianhao can this issue be closed?

lianhao:
nope, we didn't have time to resolve this in M3. should target for next release

moting9:
hostpath is easy for getting started, maybe we can have a "sample" dir in the repo,  and add diverse nfs, s3 and so on backend pv/pvc  to store model files for user reference.

PR #290, opened 2024-08-09
Enable using PV as model cache directory

yongfengdu:
Set modelUsePV to pvc name for model cache

## Description

modelUsePV support

## Issues

#128 

## Type of change

## Dependencies

## Tests

Verified using nfs as PV backend.
Setup example:
PV:
apiVersion: v1
kind: PersistentVolume
metadata:
  name: nfspv
spec:
  capacity:
    storage: 300Gi
  volumeMode: Filesystem
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain
  storageClassName: nfs
  nfs:
    path: "/data/nfspv"
    server: "192.168.0.184"
    readOnly: false

PV Claim:
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: model-volume
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: "nfs"
  resources:
    requests:
      storage: 1Mi

Test example with whisper:
helm install whisper whisper --set global.modelUseHostPath="" --set global.modelUsePV=model-volume --set replicaCount=4 --set image.tag=latest

### DIFF
--- a/helm-charts/common/speecht5/templates/deployment.yaml
+++ b/helm-charts/common/speecht5/templates/deployment.yaml
@@ -74,6 +74,9 @@ spec:
           hostPath:
             path: {{ .Values.global.modelUseHostPath }}
             type: Directory
+          {{- else if .Values.global.modelUsePV }}
+          persistentVolumeClaim:
+            claimName: {{ .Values.global.modelUsePV }}
           {{- else }}
           emptyDir: {}
           {{- end }}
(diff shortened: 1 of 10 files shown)


lianhao (approved):

lianhao:
We need to add a common README of how to use and configure PV to be used. But that could be done in another PR, since we need to figure out where to put it

daisy-ycguo (approved):
lgtm

PR #292, opened 2024-08-12
GMC: Add GPU support for GMC.

PeterYang12:
## Description

Enable NVIDIA GPU support for GMC, including sequence and switch modes. Note that switch mode may fail due to not enough GPU memory.

## Issues

## Type of change

## Dependencies

## Tests

### DIFF
--- a/microservices-connector/config/samples/chatQnA_switch_nv.yaml
+++ b/microservices-connector/config/samples/chatQnA_switch_nv.yaml
@@ -1,0 +1,124 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: gmc.opea.io/v1alpha3
+kind: GMConnector
+metadata:
+  labels:
+    app.kubernetes.io/name: gmconnector
+    app.kubernetes.io/managed-by: kustomize
+    gmc/platform: nvidia
+  name: switch
+  namespace: switch
+spec:
+  routerConfig:
+    name: router
+    serviceName: router-service
+  nodes:
+    root:
+      routerType: Sequence
+      steps:
+      - name: Embedding
+        nodeName: node1
+      - name: Reranking
+        data: $response
+        internalService:
+          serviceName: reranking-svc
+          config:
+            endpoint: /v1/reranking
+            TEI_RERANKING_ENDPOINT: tei-reranking-svc
+      - name: TeiReranking
+        internalService:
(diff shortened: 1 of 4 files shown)


microservices-connector/config/samples/chatQnA_switch_nv.yaml
```diff
@@ -0,0 +1,124 @@
...
+            config:
+              endpoint: /generate
+              MODEL_ID: Intel/neural-chat-7b-v3-3
+            isDownstreamService: true
+        - name: TgiNvidia
+          internalService:
+            serviceName: tgi-service-llama
+            config:
+              endpoint: /generate
+              MODEL_ID: Intel/neural-chat-7b-v3-3
```
irisdingbj, line 123: this is supposed to be the llama model

microservices-connector/usage_guide.md
```diff
@@ -39,7 +43,7 @@ kubectl create deployment client-test -n chatqa --image=python:3.8.13 -- sleep i
 **Access the pipeline using the above URL from the client pod**
 
 ```bash
-export CLIENT_POD=$(kubectl get pod  -l app=client-test -o jsonpath={.items..metadata.name})
+export CLIENT_POD=$(kubectl get pod -n chatqa  -l app=client-test -o jsonpath={.items..metadata.name})
```
irisdingbj, line 46: good catch!

microservices-connector/config/samples/chatQnA_switch_nv.yaml
PeterYang12: Thank you for your review. Here is a workaround. If using the llama model, one GPU card may fail to launch two instance due to few GPU memory. I am ok to keep the llama model here, what do you think?

irisdingbj: this is just to work around for current NV machine, for the example we want to provide to end user, it is better we use a meaningful example, like keep the llama model.

PeterYang12: > this is just to work around for current NV machine, for the example we want to provide to end user, it is better we use a meaningful example, like keep the llama model.

Fixed

irisdingbj (approved):

daisy-ycguo (approved):
lgtm

PeterYang12:
Thank you, Iris and Daisy. :)

Issue #276, opened 2024-08-07
helm-chart/common/chatqna-ui refers to docker image opea/chatqna-conversation-ui but this does not exist on docker hub


SeanCondon:
Hi I am in the ITEP development group along with @kushal2705 
I'm trying out the chatqna-conversation-ui but it refers to this unknown version - I've tried the 0.7 and latest and while the UI opens it cannot send a POST request to the backend

lianhao:
We assumed the public container image v0.8 on docker hub should be available soon after the v0.8 release(released at end of July), unfortunately it's not the case.

As for your UI specific issue, because of the issue opea-project/GenAIExamples#493 and opea-project/GenAIExamples#518, you have to manually build the UI docker image by yourselves at this time. 

@Ruoyu-y Could you please provide some workaround guide on how to make UI working before the above 2 issues are resolved? Thanks!

SeanCondon:
I created a PR #305 that adjusts the helm chart to make the backend configurable at deploy time.
It creates a reverse proxy in nginx that points to the backend. The UI then points to this reverse proxy using a relative path e.g. `/chatqna/v1/chatqna` 

In this case I build the UI with
```bash
docker build --platform=linux/amd64 --no-cache -t opea/chatqna-conversation-ui:itep-build --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy --build-arg BACKEND_SERVICE_ENDPOINT=/chatqna/v1/chatqna --build-arg DATAPREP_SERVICE_ENDPOINT=/chatqna-data-prep/v1/dataprep -f ./docker/Dockerfile.react .
```

lianhao:
I believe the docker image is published and CORS issue is fixed

PR #305, opened 2024-08-14
chatqna-ui helm chart - nginx reverse proxy

SeanCondon:
## Description

This PR adds an nginx reverse proxy to the ChatQnA UI helm chart which allows the path to the `chatqna` pod and the `data-prep` pod to be configured at deploy time. This also has the benefit that it removes CORS errors

## Issues

#276 

## Type of change

## Dependencies

## Tests


lianhao:
@SeanCondon please fix the DCO failure. You may need to 'git commit -s' to add the sign-off-by line in your commit message

lianhao:
I think we also need to disable the 'helm test' for this case(i.e. delete the testpod definition). Since we can't have a backward dependency from ui to backend in this repo layout.

helm-charts/common/chatqna-ui/templates/nginx-conf.yaml
```diff
@@ -0,0 +1,39 @@
...
+        try_files $uri $uri/ /index.html =404;
+
+        location ~* \.(gif|jpe?g|png|webp|ico|svg|css|js|mp4|woff2)$ {
+            expires 1d;
+        }
+      }
+
+      location {{ .Values.BACKEND_SERVICE_ENDPOINT }} {
+        proxy_pass {{ .Values.backend.scheme }}://{{ .Release.Name }}-{{ .Values.backend.name }}:{{ .Values.backend.port }}/;
+      }
```
yongfengdu, line 34: Would prefer to keep the meaning of BACKEND_SERVICE_ENDPOINT as the same, because docker-compose is also using it.
Also, the {{ .Release.Name }} is not necessary part of the ENDPOINT, since it's possible chatqna-ui and backend have different release name. We need to give user options to change that.
Hardcode the location shouldn't be a problem, if that is, let's define a new variable to do that.
Recommend this way:
      location "/chatqna/" {
        {{- if .Values.BACKEND_SERVICE_ENDPOINT }}
        proxy_pass {{ .Values.BACKEND_SERVICE_ENDPOINT }}
        {{- else }}
        proxy_pass {{ .Values.backend.scheme }}://{{ .Release.Name }}-{{ .Values.backend.name }}:{{ .Values.backend.port }}/;
      }

helm-charts/common/chatqna-ui/templates/nginx-conf.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+      location {{ .Values.DATAPREP_SERVICE_ENDPOINT }} {
```
yongfengdu, line 36: ditto

helm-charts/common/chatqna-ui/values.yaml
```diff
@@ -62,11 +62,21 @@ tolerations: []
...
+  port: 8888
+
+dataprep:
+  scheme: http
+  name: data-prep
+  port: 6007
+
 # chatQnA Mega service URL, e.g. http://<service-name>:<port>
-BACKEND_SERVICE_ENDPOINT: ""
+BACKEND_SERVICE_ENDPOINT: "/chatqna/"
```
yongfengdu, line 76: No change of this variable's meaning as said above.

helm-charts/common/chatqna-ui/values.yaml
```diff
@@ -62,11 +62,21 @@ tolerations: []
...
 
 # data preparation service URL, http://<service-name>:<port>
-DATAPREP_SERVICE_ENDPOINT: ""
+DATAPREP_SERVICE_ENDPOINT: "/chatqna-data-prep/"
```
yongfengdu, line 79: ditto

yongfengdu (commented):
It took me sometime to figure out how to use this, so a README is welcome for the UI.

Ruoyu-y:
Regarding this, I saw @jaswanth8888 has fixed the issue https://github.com/opea-project/GenAIExamples/issues/493 on defining the urls during build. But it does not clean up the CORS issue. Maybe can @SeanCondon and @jaswanth8888 get aligned here? So that we would have same logic for the UI in both docker and kubernetes case?

SeanCondon:
> I think we also need to disable the 'helm test' for this case(i.e. delete the testpod definition). Since we can't have a backward dependency from ui to backend in this repo layout.

This is correct - it is a side effect of the strict checking of downstream dependencies by nginx pod.

SeanCondon:
> Thanks @SeanCondon for creating this PR. Regarding this, I saw @jaswanth8888 has fixed the issue [opea-project/GenAIExamples#493](https://github.com/opea-project/GenAIExamples/issues/493) on defining the urls during build. But it does not clean up the CORS issue. Maybe can @SeanCondon and @jaswanth8888 get aligned here? So that we would have same logic for the UI in both docker and kubernetes case?

@Ruoyu-y I have created a [PR](https://github.com/opea-project/GenAIExamples/pull/616) in `GenAIExamples/ChatQnA` that addresses this in the docker-compose case. I will come back and adjust this PR is that one is merged.

lianhao:
@SeanCondon @Ruoyu-y has merged a PR #377 to add the nginx reverse proxy at chatqna e2e level. So I'm wondering if this PR is still needed or not?

SeanCondon:
> @SeanCondon @Ruoyu-y has merged a PR #377 to add the nginx reverse proxy at chatqna e2e level. So I'm wondering if this PR is still needed or not?

Thanks for letting me know about #377 - I had not seen it. If the problem is solved then this change is not needed. Please feel free to reopen if needed.  cc @Ruoyu-y. 

Please see [https://github.com/opea-project/GenAIExamples/pull/616](https://github.com/opea-project/GenAIExamples/pull/616) where I address this for `docker compose`.

PR #306, opened 2024-08-15
Update helm chart readme

### LINKED ISSUE #128: Support production level volume types for model data

yongfengdu:
## Description

Update helm chart readme, include how to use PersistentVolume as model cache.

## Issues

Closed #128 

## Type of change

## Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@ Once you have deployed GMC in your Kubernetes cluster, you can deploy any of the
 
 To deploy GenAIExamples to Kubernetes using helm charts, you need [Helm](https://helm.sh/docs/intro/install/) installed on your machine.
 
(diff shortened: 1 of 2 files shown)


daisy-ycguo (approved):
lgtm

lianhao (approved):

PR #315, opened 2024-08-19
fix: update doc for authN-authZ with oauth

Ruoyu-y:
## Description

Some fix for authN and authZ with oauth2-proxy + OIDC
* remove plain-text cookie_secret from config file
* fix minor issues in doc
* remove redundant info in doc

## Issues

## Type of change

- [*] Bug fix (non-breaking change which fixes an issue)

## Dependencies

`n/a`

## Tests

`n/a`


daisy-ycguo (approved):
Thank you.

PR #355, opened 2024-08-27
mainfest: introduce release namespace object to endpoint

KarChunT:
## Description

Introduce the **release namespace object** to the **endpoint**, as the current one only supports **default** namespace.

## Issues

`n/a`

## Type of change

## Dependencies

`n/a`

## Tests

Tested the ChatQNA with a different namespace, able to work perfectly.
image


lianhao (commented):
There is no need to add namespace suffix to URL, because k8s will automatically append that during in-cluster DNS resolution

PR #383, opened 2024-09-03
GMC: add UT for reconcile filters

KfreeZ:
## Description

1. add UT for reconcile filters
2. in order to do the UT, extract the filter out of SetupWithManager

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

CI/CD covers the regression tests
controller coverage increased 
ok      github.com/opea-project/GenAIInfra/microservices-connector/internal/controller  22.452s coverage: 78.2% of statements


.github/workflows/scripts/e2e/gmc_xeon_test.sh
```diff
@@ -137,7 +137,7 @@ function validate_audioqa() {
     pods_count=$(kubectl get pods -n $AUDIOQA_NAMESPACE -o jsonpath='{.items[*].metadata.name}' | wc -w)
     # expected_ready_pods, expected_external_pods, expected_total_pods
     # pods_count-1 is to exclude the client pod in this namespace
-    check_gmc_status $AUDIOQA_NAMESPACE 'audioqa' $((pods_count-1)) 0 7
```
zhlsunshine, line 140: why do you use precise number instead of `pods_count` ?

zhlsunshine, line 140: Hi @KfreeZ, why do you use precise number instead `pods_count` ?

KfreeZ: hi @zhlsunshine  because the pod count is not accurate sometimes when terminating pod is not cleared but counted

zhlsunshine: Okay, I through that it should be okay after calling `wait_until_all_pod_ready`, if yes, could we just fix the bug for this function?

KfreeZ: the latest wait_until_all_pod_ready has excluded the "Terminating" pods, so there is possibility that the pods are still existing as "Terminating" pods after wait_until_all_pod_ready. Yes I can also count the pods by excluding the "Terminating" pods, but for these test case, the ready counts and total counts are fixed numbers, for chatqa for instances, after a deploy, "8/0/8" is exactly we expected.

zhlsunshine: okay, then let's exclude the `Terminating` pods when calculating the `pods_count`.

zhlsunshine (approved):
LGTM

lianhao:
Do we have explanation of this CI test failure: https://github.com/opea-project/GenAIInfra/actions/runs/10697764146/job/29655792762

Issue #394, opened 2024-09-04
GMC manager panic during CI GMC e2e test


lianhao:
Using the latest code(commit id: `59ce50506`), during the GMC e2e CI test, I found the GMC manager restarted several times due to panic:

```
$ kubectl -n opea-system get pod
NAME                             READY   STATUS    RESTARTS        AGE
gmc-controller-f67965865-nz89w   1/1     Running   2 (3m36s ago)   7m13s
```

```
$ kubectl -n opea-system logs gmc-controller-f67965865-nz89w -p
2024-09-04T12:56:34Z    INFO    WebhookConfig   Creating or updating the validatingwebhookconfiguration {"webhook name": "validating-webhook-configuration"}
2024-09-04T12:56:34Z    INFO    WebhookConfig   Updated the validatingWebhookConfiguration      {"webhookConfigName": "validating-webhook-configuration"}
2024-09-04T12:56:34Z    INFO    controller-runtime.builder      skip registering a mutating webhook, object does not implement admission.Defaulter or WithDefaulter wasn't called       {"GVK": "gmc.opea.io/v1alpha3, Kind=GMConnector"}
2024-09-04T12:56:34Z    INFO    controller-runtime.builder      Registering a validating webhook        {"GVK": "gmc.opea.io/v1alpha3, Kind=GMConnector", "path": "/validate-gmc-opea-io-v1alpha3-gmconnector"}
2024-09-04T12:56:34Z    INFO    controller-runtime.webhook      Registering webhook     {"path": "/validate-gmc-opea-io-v1alpha3-gmconnector"}
2024-09-04T12:56:34Z    INFO    setup   starting manager
2024-09-04T12:56:34Z    INFO    controller-runtime.metrics      Starting metrics server
2024-09-04T12:56:34Z    INFO    starting server {"kind": "health probe", "addr": "[::]:8081"}
2024-09-04T12:56:34Z    INFO    controller-runtime.metrics      Serving metrics server  {"bindAddress": ":8080", "secure": false}
2024-09-04T12:56:34Z    INFO    controller-runtime.webhook      Starting webhook server
2024-09-04T12:56:34Z    INFO    controller-runtime.webhook      Serving webhook server  {"host": "", "port": 9443}
2024-09-04T12:56:34Z    INFO    Starting EventSource    {"controller": "gmconnector", "controllerGroup": "gmc.opea.io", "controllerKind": "GMConnector", "source": "kind source: *v1alpha3.GMConnector"}
[...]

PR #395, opened 2024-09-04
fix GMC crashes in e2e 

KfreeZ:
## Description

fix for #394 

**Root cause analysis:**

this issue is introduced by #383 , in this PR, I moved the deployment filters, but missed the 
```
		CreateFunc: func(e event.CreateEvent) bool {
			return false
		}, DeleteFunc: func(e event.DeleteEvent) bool {
			return false
		},
```
without these two, the creation/deletion of the deployment triggers the reconcile, 

the sequence to hit this issue is like: delete GMC pipeline -> trigger deployment deletion ->trigger graph status update, but graph is already deleted ->panic.

**solution：**

- restore the create/delete filters
- add protection code

## Issues

#394 

## Type of change

## Dependencies

n/a

## Tests

CI/CD cover the tests

### DIFF
--- a/microservices-connector/internal/controller/gmconnector_controller.go
+++ b/microservices-connector/internal/controller/gmconnector_controller.go
@@ -825,12 +825,12 @@ func isDeploymentStatusChanged(e event.UpdateEvent) bool {
 	}
 
 	if len(newDeployment.OwnerReferences) == 0 {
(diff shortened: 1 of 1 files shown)


irisdingbj (approved):

mkbhanda (approved):
LGTM -- thank you for the detailed root cause analysis @KfreeZ

PR #402, opened 2024-09-05
Add TGI additional options

yongfengdu:
Add user configurable shm_size support.
Add interface for additional TGI cli parameters.

## Description

## Issues

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/common/tgi/values.yaml
+++ b/helm-charts/common/tgi/values.yaml
@@ -11,6 +11,11 @@ horizontalPodAutoscaler:
   maxReplicas: 6
 
 port: 2080
+shmSize: 1Gi
+
+# Set extraCmdArgs if you need to pass additional parameters to TGI for performance
+# Refer to https://huggingface.co/docs/text-generation-inference/en/reference/launcher for more options.
+# extraCmdArgs: ["--dtype","bfloat16"]
 
 image:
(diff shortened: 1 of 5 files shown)


lianhao (approved):

helm-charts/common/tgi/values.yaml
```diff
@@ -11,6 +11,11 @@ horizontalPodAutoscaler:
   maxReplicas: 6
 
 port: 2080
+shmSize: 1Gi
+
+# Set extraCmdArgs if you need to pass additional parameters to TGI for performance
+# Refer to https://huggingface.co/docs/text-generation-inference/en/reference/launcher for more options.
+# extraCmdArgs: [""]
```
eero-t, line 18: Could you add also (commented out) example of how to set _multiple_ args?

For example `--dtype bfloat16` (for newer CPUs).

daisy-ycguo (approved):
lgtm

eero-t:
It seems that this is needed only when model requires more memory than fits to given device, i.e. it needs to be sharded over multiple devices using something like deepspeed?
 
This means that it would need to be in the relevant top-level <device>-values.yaml files, where the model and device allocations are specified.  I.e. model, allocation for number of devices (>1) needed for it, matching SHM size, and TGI sharding options all need to go hand-in-hand.

PR #405, opened 2024-09-06
Add topologySpreadConstraints support

### LINKED ISSUE #401: [Requirement]Request for Adding TopologySpreadConstraints to ChatQnA Helm Chart
Is it possible to add topologySpreadConstraints to the ChatQnA Helm chart? This would help ensure that when we scale the pods across multiple nodes, they are scheduled evenly across those nodes.
Below is an example YAML snippet for reference:

```
      topologySpreadConstraints:
      - maxSkew: 1
        topologyKey: kubernetes.io/hostname
        whenUnsatisfiable: ScheduleAnyway
        labelSelector:
               matchLabels:
                 app: tei-embedding-service-deploy
```

yongfengdu:
User can set evenly_distributed true to have the pods distribute evenly.

## Issues

Closed #401 

## Type of change

### DIFF
--- a/helm-charts/chatqna/templates/deployment.yaml
+++ b/helm-charts/chatqna/templates/deployment.yaml
@@ -84,3 +84,12 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.evenly_distributed }}
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: kubernetes.io/hostname
+          whenUnsatisfiable: ScheduleAnyway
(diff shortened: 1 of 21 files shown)


lianhao:
Should we have this documented somewhere?

lianhao (approved):
We may need to document this in top level helm-chart README?

leslieluyu (approved):

eero-t:
> We may need to document this in top level helm-chart README?

Definitely, this is one more alternative for fixing the deployments, in addition to using correct resources values, or using affinity rules (to do similar thing as this does).

PR #451, opened 2024-09-23
Update the tgi service images for gaudi

zhlsunshine:
## Description

Update the tgi service images for both xeon and gaudi.

## Issues

`n/a`.

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

`n/a`.

## Tests

`n/a`.

PR #454, opened 2024-09-23
Support alternative metrics on accelerated TGI / TEI instances

eero-t:
## Description

Changes:
* Provide `accelDevice` Helm variables also in other charts using accelerated TGI / TEI
* Use different custom / HPA metrics depending on whether TGI / TEI is accelerated

Queue size metric is a bit better for scaling TGI Gaudi instances than the one used for Xeon scaling, and queries for it are simpler.  HPA rule updates act also as examples on use of different HPA algorithm types (`Value` vs. `AverageValue`).

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

Did manual testing of HPA scaling of 1-4 TGI Gaudi instances (4 = maxReplicas value), using different TGI-provided metrics, and batch sizes.

PS. none of the TGI / TEI metrics seem that good for scaling, their values mostly indicate just whether given component is stressed, not _how_ slow the resulting (e.g. ChatQnA) queries are.   Meaning that HPA scaling factor causes scale up when they get loaded, and it drops extra instances when load goes almost completely away, but that scaling can be quite far from a really optimal one.

Something like https://github.com/opea-project/GenAIExamples/issues/391 (and fine-tuning the HPA values for given model, data type, OPEA version etc) would be needed for more optimal scaling.

### DIFF
--- a/helm-charts/chatqna/templates/custom-metrics-configmap.yaml
+++ b/helm-charts/chatqna/templates/custom-metrics-configmap.yaml
@@ -12,54 +12,75 @@ metadata:
     app.kubernetes.io/name: prometheus-adapter
 data:
   config.yaml: |
-      rules:
-      {{- if .Values.tgi.horizontalPodAutoscaler.enabled }}
-      # check metric with:
-      # kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1/namespaces/default/service/*/{{ include "tgi.metricPrefix" .Subcharts.tgi }}_request_latency | jq
-      #
-      - seriesQuery: '{__name__="tgi_request_inference_duration_sum",service="{{ include "tgi.fullname" .Subcharts.tgi }}"}'
-        # Average request latency from TGI histograms, over 1 min
-        # (0.001 divider add is to make sure there's always a valid value)
-        metricsQuery: 'rate(tgi_request_inference_duration_sum{service="{{ include "tgi.fullname" .Subcharts.tgi }}",<<.LabelMatchers>>}[1m]) / (0.001+rate(tgi_request_inference_duration_count{service="{{ include "tgi.fullname" .Subcharts.tgi }}",<<.LabelMatchers>>}[1m]))'
-        name:
-          matches: ^tgi_request_inference_duration_sum
-          as: "{{ include "tgi.metricPrefix" .Subcharts.tgi }}_request_latency"
-        resources:
-          # HPA needs both namespace + suitable object resource for its query paths:
-          # /apis/custom.metrics.k8s.io/v1beta1/namespaces/default/service/*/tgi_request_latency
-          # (pod is not suitable object type for matching as each instance has different name)
-          overrides:
-            namespace:
-              resource: namespace
-            service:
-              resource: service
-      {{- end }}
-      {{- if .Values.teirerank.horizontalPodAutoscaler.enabled }}
-      - seriesQuery: '{__name__="te_request_inference_duration_sum",service="{{ include "teirerank.fullname" .Subcharts.teirerank }}"}'
-        # Average request latency from TEI histograms, over 1 min
-        metricsQuery: 'rate(te_request_inference_duration_sum{service="{{ include "teirerank.fullname" .Subcharts.teirerank }}",<<.LabelMatchers>>}[1m]) / (0.001+rate(te_request_inference_duration_count{service="{{ include "teirerank.fullname" .Subcharts.teirerank }}",<<.LabelMatchers>>}[1m]))'
-        name:
-          matches: ^te_request_inference_duration_sum
-          as: "{{ include "teirerank.metricPrefix" .Subcharts.teirerank }}_request_latency"
-        resources:
-          overrides:
-            namespace:
-              resource: namespace
-            service:
-              resource: service
-      {{- end }}
-      {{- if .Values.tei.horizontalPodAutoscaler.enabled }}
-      - seriesQuery: '{__name__="te_request_inference_duration_sum",service="{{ include "tei.fullname" .Subcharts.tei }}"}'
-        # Average request latency from TEI histograms, over 1 min
-        metricsQuery: 'rate(te_request_inference_duration_sum{service="{{ include "tei.fullname" .Subcharts.tei }}",<<.LabelMatchers>>}[1m]) / (0.001+rate(te_request_inference_duration_count{service="{{ include "tei.fullname" .Subcharts.tei }}",<<.LabelMatchers>>}[1m]))'
-        name:
-          matches: ^te_request_inference_duration_sum
-          as: "{{ include "tei.metricPrefix" .Subcharts.tei }}_request_latency"
-        resources:
-          overrides:
-            namespace:
-              resource: namespace
-            service:
-              resource: service
-      {{- end }}
+    rules:
+    {{- if .Values.tgi.horizontalPodAutoscaler.enabled }}
+    # check metric with:
+    # kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1/namespaces/default/service/*/<metric> | jq
+    #
+    {{- if .Values.tgi.accelDevice }}
+    - seriesQuery: '{__name__="tgi_queue_size",service="{{ include "tgi.fullname" .Subcharts.tgi }}"}'
+      # TGI instances queue_size sum
+      metricsQuery: 'sum by (namespace,service) (tgi_queue_size{service="{{ include "tgi.fullname" .Subcharts.tgi }}",<<.LabelMatchers>>})'
+      name:
+        matches: ^tgi_queue_size
+        as: "{{ include "tgi.metricPrefix" .Subcharts.tgi }}_queue_size_sum"
+    {{- else }}
+    - seriesQuery: '{__name__="tgi_request_inference_duration_sum",service="{{ include "tgi.fullname" .Subcharts.tgi }}"}'
+      # Average request latency from TGI histograms, over 1 min
+      # (0.001 divider add is to make sure there's always a valid value)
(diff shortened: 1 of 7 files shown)


lianhao:
@eero-t could you double check the CI failure on xeon with `cpu-values.yaml`? You may try to reproduce and debug locally by `helm install chatqna . -f cpu_values.yaml` to install chatqna on xeon platform, and  `helm test chatqna` to figure out if something is failing

eero-t:
Failure in CI CPU values file test is for the TGI service response check: https://github.com/opea-project/GenAIInfra/actions/runs/10991943008/job/30556825567?pr=451

This PR changes only HPA files and CI skips HPA testing, so the reason for the failure is some change done *before* this PR.

**What CPU values file does**

CPU values file overrides deployment probe timings and resources from defaults: https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/tgi/values.yaml#L58

So that ChatQnA service can be scaled better without failing, as scheduler knows how much resources they need, and kubelet uses higher QoS class for them: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/

Increased probe timings cannot be the cause for failure, and memory limits should be large enough, not to be an issue either.  CPU limits are quite tight though (8 CPUs) to allow running multiple TGI instances on same node, especially with NRI policies.

**Reason for fail**

CI test checks TGI log output, but appears to be doing it too soon.

CPU limits can can explain slowdown from default (no resources specified), when compared to situation where only single TGI instance is running / node has plenty of resources free.

But CPU values file passed earlier, so something else has changed since it was merged, either in CI setup, or the TGI setup (other than values specified in CPU file).

**Potential workarounds**

Things that can be done (in some other PR):
* CI:
  * increase time for checking TGI log, or
  * return CI setup to earlier state where TGI startup was faster
* CPU values file:
  * increase (TGI) CPU resources

yongfengdu:
Another possibility is the latest image updated from docker hub, which might cause different behaviors.
The returned output seems correct, however, helm test for llm-uservice failed.
Need further investigation on this.

eero-t:
Reviewed tests that were run for _all_ PRs that were merged since `cpu-values.yaml` was added in #386.

=> this and #451 are _first_ PRs for which CI tested `cpu-values.yaml`, despite there being several merged PRs which could have broken things for it already earlier, as they updated CPU / Xeon related files in ChatQnA and its dependencies (e.g. #444).

Additionally, when looking at the passing Helm E2E tests in that merged #386 PR, e.g. for "xeon, values, common": https://github.com/opea-project/GenAIInfra/actions/runs/10800184120/job/29957611425?pr=386

There are lot of "Couldn't connect to server" errors (e.g. for data-prep), but the test passed despite failures in functionality not touched by that PR.

Then when looking at same passing test in later PRs, tests do not log  any more what they are doing: https://github.com/opea-project/GenAIInfra/actions/runs/10940026240/job/30371535827?pr=444

I.e. it seems that CI tests cannot be trusted to guarantee any of the functionality working, or catching when that functionality actually broke.

=> Somebody (not me) needs to bisect repo content in the CI environment to see when things actually broke.

eero-t:
Additionally, E2E tests pass the tests even if log file content was wrong, if the log file had wrong name, see:
https://github.com/opea-project/GenAIInfra/blob/main/.github/workflows/_helm-e2e.yaml#L142

lianhao:
> Additionally, when looking at the passing Helm E2E tests in that merged #386 PR, e.g. for "xeon, values, common": https://github.com/opea-project/GenAIInfra/actions/runs/10800184120/job/29957611425?pr=386
> 
> There are lot of "Couldn't connect to server" errors (e.g. for data-prep), but the test passed despite failures in functionality not touched by that PR.

This is a workaround in the testpod to avoid some opea services' probe health check status doesn't got populated in the K8S service availability, sleep for a short period of time between 'helm install --wait' and 'helm test' could be another option.

> Then when looking at same passing test in later PRs, tests do not log any more what they are doing: https://github.com/opea-project/GenAIInfra/actions/runs/10940026240/job/30371535827?pr=444

What does the test do is in each helm chart's `template/tests/test-pod.yaml`. The log only shows the test output.

> I.e. it seems that CI tests cannot be trusted to guarantee any of the functionality working, or catching when that functionality actually broke.

lianhao:
ok, I root caused this issue. [GenAIComps PR 608](https://github.com/opea-project/GenAIComps/pull/608) plus cpu-values significantly increases the test run time, which trigger the default 5m timeout of `helm test`. Please wait for PR #456 to land-in first

eero-t:
> ok, I root caused this issue. [GenAIComps PR 608](https://github.com/opea-project/GenAIComps/pull/608) plus cpu-values significantly increases the test run time, which trigger the default 5m timeout of `helm test`. Please wait for PR #456 to land-in first

How much `cpu-values.yaml` increased the test time with PR 608?   If it's really significant, it would be better to raise (also) the CPU limits in it.

(The values I've used may be too small.  At that time I had only couple of Xeon nodes for scaling testing, and smaller values allowed stuffing more instances to a single Xeon node.  However, I used also NRI to isolate pods better, which is not the default k8s setup...)

lianhao (approved):

eero-t:
> This is a workaround in the testpod to avoid some opea services' probe health check status doesn't got populated in the K8S service availability quickly enough.

Tests should be predictable and fail on errors, not paper over them.
 
> Sleeping for a short period of time between 'helm install --wait' and 'helm test' could be another option.

Script should wait until pod is ready to accept input before trying to use it, or fail test if wait timeouts.

For example, this fails unless there are matching TGI pods, and all of them are in Ready state, within given timeout:
```
if ! kubectl wait --for=condition=Ready --timeout=60s --selector=app.kubernetes.io/name=tgi pod; then
  echo "ERROR: TGI pods not ready within 60s"
  exit 1
fi
```

> What does the test do is in each helm chart's `template/tests/test-pod.yaml`. The log only shows the test output.

That's not really good enough.  One would need to dig out all the relevant test template files from the given PR branch, go through all relevant values files and deduct results for all variables & conditionals in the template.   That's way too much work and error prone.

Tests should log what they're actually testing, otherwise they're not worth much.  Without logging it's nightmare to verify them testing everything that's relevant (and preferably skip completely irrelevant things), and debug failures.

eero-t:
> > This is a workaround in the testpod to avoid some opea services' probe health check status doesn't got populated in the K8S service availability quickly enough.
> 
> Tests should be predictable and fail on errors, not paper over them.
> 
> > Sleeping for a short period of time between 'helm install --wait' and 'helm test' could be another option.
> 
> Script should wait until pod is ready to accept input before trying to use it, or fail test if wait timeouts.
> 
> For example, this fails unless there are matching TGI pods, and all of them are in `Ready` state, within given timeout:
> 
> ```
> if ! kubectl wait --for=condition=Ready --timeout=60s --selector=app.kubernetes.io/name=tgi pod; then
>   echo "ERROR: TGI pods not ready within 60s"
>   exit 1
> fi
> ```
> 
> > What does the test do is in each helm chart's `template/tests/test-pod.yaml`. The log only shows the test output.
> 
> That's not really good enough. One would need to dig out all the relevant test template files from the given PR branch, go through all relevant values files and deduct results for all variables & conditionals in the template. That's way too much work and error prone.
> 
> Tests should log what they're actually testing, otherwise they're not worth much. Without logging it's nightmare to verify them correctly testing everything that's relevant (and preferably skip completely irrelevant things), and debug failures.

eero-t:
> > > This is a workaround in the testpod to avoid some opea services' probe health check status doesn't got populated in the K8S service availability quickly enough.
> > 
> > 
> > Tests should be predictable and fail on errors, not paper over them.
> > > Sleeping for a short period of time between 'helm install --wait' and 'helm test' could be another option.
> > 
> > 
> > Script should wait until pod is ready to accept input before trying to use it, or fail test if wait timeouts.
> > For example, this fails unless there are matching TGI pods, and all of them are in `Ready` state, within given timeout:
> > ```
> > if ! kubectl wait --for=condition=Ready --timeout=60s --selector=app.kubernetes.io/name=tgi pod; then
> >   echo "ERROR: TGI pods not ready within 60s"
> >   exit 1
> > fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > What does the test do is in each helm chart's `template/tests/test-pod.yaml`. The log only shows the test output.
> > 
> > 
> > That's not really good enough. One would need to dig out all the relevant test template files from the given PR branch, go through all relevant values files and deduct results for all variables & conditionals in the template. That's way too much work and error prone.
> > Tests should log what they're actually testing, otherwise they're not worth much. Without logging it's nightmare to verify them correctly testing everything that's relevant (+ preferably skipping completely irrelevant things), and debug failures.

lianhao:
> Script should wait until pod is ready to accept input before trying to use it, or fail test if wait timeouts.
> 
> For example, this fails unless there are matching TGI pods, and all of them are in `Ready` state, within given timeout:
> 
> ```
> if ! kubectl wait --for=condition=Ready --timeout=60s --selector=app.kubernetes.io/name=tgi pod; then
>   echo "ERROR: TGI pods not ready within 60s"
>   exit 1
> fi
> ```
the current `helm install --wait` actually have already done exactly the same as the above, but for all pods I think,

eero-t:
> the current `helm install --wait` actually have already done exactly the same as the above, but for all pods I think,

Oh, that waits also for Ready state. Yes, that would be the correct thing to do:
```
$ helm install -h
...
--wait if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout
```

yongfengdu (approved):
It looks to me we shouldn't use "helm install -f cpu-values.yaml" for test since the cpu-values.yaml is for HPA usage.
Maybe we can consider rename it to cpu-hpa-values.yaml or something.

Issue #281, opened 2024-08-08
Extend support for chat history and prompt registry comps under GMC


hteeyeoh:
ChatHistory: https://github.com/yogeshmpandey/GenAIComps/tree/main/comps/chathistory/mongo
PromptRegistry: https://github.com/yogeshmpandey/GenAIComps/tree/main/comps/prompt_registry/mongo

zhlsunshine:
Hi @hteeyeoh, thanks for your requirement, basically, the helm chart should support these 2 components first, then we need to add them into ChatQnA example, including implement both the helm chart and GMC, so I also @yongfengdu and @lianhao here as well.
We need to talk about these components and related examples to decide the support. Thanks.

hteeyeoh:
Thanks @zhlsunshine . Appreciate that if @yongfengdu or @lianhao can give ETA on this if possible so that our team can plan accordingly. Thank you.

lianhao:
We're busy with 1.0 release now. Will loop you in when we're doing planning after 1.0 release.

hteeyeoh:
Sure. No problem. Thank you. Keep us posted.

PR #458, opened 2024-09-26
Add helm chart for 3 components

yongfengdu:
ChatHistory/PromptRegistry and mongodb.

## Description

Add helm charts for 3 components

## Issues

#281 

## Type of change

## Dependencies

## Tests

Local build opea/promptregistry-mongo and opea/chathistory-mongo-server.
helm install and helm test

### DIFF
--- a/helm-charts/common/mongodb/templates/deployment.yaml
+++ b/helm-charts/common/mongodb/templates/deployment.yaml
@@ -1,0 +1,76 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "mongodb.fullname" . }}
+  labels:
+    {{- include "mongodb.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "mongodb.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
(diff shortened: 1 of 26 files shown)


helm-charts/common/mongodb/templates/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+          volumeMounts:
+            - mountPath: /data
+              name: data-volume
+            - mountPath: /tmp
+              name: tmp
+          ports:
+            - name: mongodb
+              containerPort: 27017
+              protocol: TCP
+          startupProbe:
```
lianhao, line 45: maybe we should move this to values.yaml, just incase the enduser may need to change it in the future

lianhao:
there seems no chat-history and prompt image in CI's container repo.

helm-charts/common/mongodb/templates/deployment.yaml
yongfengdu: 27017 is a well known port for mongodb, and since this port belongs to the pod itself, I don't think user need to change this value.
If user want to change the mongdb service's port, he can use the port defined for service:
     - port: {{ .Values.service.port }}

yongfengdu:
> there seems no chat-history and prompt image in CI's container repo.
It's available now in the opea repo, I'll check how to fix the CI issue.
https://github.com/opea-project/GenAIComps/issues/718

lianhao (approved):

daisy-ycguo (approved):

PR #473, opened 2024-10-15
Refactor CI scripts to support more components

### LINKED ISSUE #471: [CI/helm] Enhancement of CI for helm charts
Current CI for helm charts has 2 problems:
1). It use autodependency.enabled flag to handle the dependency, it works fine with one set of dependency, but with more flavors supported, this is hard to manage.
For Example:
data-prep depends on 2 services: Embedding and VectorDB.
For an AI app, it's common these services are shared with retriever-usvc, so it's the App's responsibility to pass the Embedding/VectorDB Endpoint to data-prep. However for CI to test the data-prep service itself, we'll need to bring up our own Embedding and VectorDB, autodependency.enabled provide the flexibility of deploying dependent services.
The original version only support TEI and Redis, so it works fine, but with more flavors support(Milvus as vectorDB, mosec as embedding endpoint), autodependency is not flexible enough to cover this, we need to fix it.
2). Current CI will run test on all different configurations(*values.yaml).
CI will run tests on all the values.yaml it found for the helm charts under test, and skip undesired one in the test script.
This might test more configurations than we expected. (cpu-values.yaml etc), and some values.yaml are added for reference only.
The better way is to explicitly specify CI covered cases, with well defined naming rules.
For example, we can use *-ci-values.yaml to let the CI know this values file is for CI to use.

yongfengdu:
## Description

Refactor the CI scripts.

## Issues

Closed #471 

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/CI.md
+++ b/helm-charts/CI.md
@@ -1,0 +1,25 @@
+# CI guidelines for helm charts
+
+## Table of Contents
+
+- [Infra Setup](#infra-setup)
+- [Add new test case](#add-new-test-case)
+
+## Infra Setup
+
+Helm charts are organized as:
+1). E2E examples are directly located under helm-charts/ directory
+2). Charts for components are located under helm-charts/common directory.
+
+We have Xeon servers with Gaudi accelerator card, which covers CPU and Gaudi accelerated workloads. For coverage of other configurations/features, it's case by case.
+
+CI tests values files which names are listed in `ci-*values.yaml` files.
+
+## Add new test case
+
+Add a new values file following the naming rules: `ci-*values.yaml`.
+If the values file for CI is identical to another one, a symbolic link can be used.
+
+See following files for the test scripts:
+https://github.com/opea-project/GenAIInfra/blob/main/.github/workflows/pr-chart-e2e.yaml
+https://github.com/opea-project/GenAIInfra/blob/main/.github/workflows/_helm-e2e.yaml
--- a/helm-charts/chatqna/ci-gaudi-values.yaml
+++ b/helm-charts/chatqna/ci-gaudi-values.yaml
@@ -1,0 +1,1 @@
+gaudi-values.yaml
--- a/helm-charts/chatqna/ci-guardrails-gaudi-values.yaml
+++ b/helm-charts/chatqna/ci-guardrails-gaudi-values.yaml
@@ -1,0 +1,1 @@
+guardrails-gaudi-values.yaml
--- a/helm-charts/chatqna/ci-guardrails-values.yaml
+++ b/helm-charts/chatqna/ci-guardrails-values.yaml
@@ -1,0 +1,1 @@
+guardrails-values.yaml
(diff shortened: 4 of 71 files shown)


helm-charts/CI.md
```diff
@@ -0,0 +1,26 @@
+# CI guidelines for helm charts
+
+## Table of Contents
+
+- [Infra Setup](#infra-setup)
+- [Add new test case](#add-new-test-case)
+
+## Infra Setup
+
+Helm charts are orginized as:
```
eero-t, line 10:
```suggestion
Helm charts are organized as:
```

helm-charts/CI.md
```diff
@@ -0,0 +1,26 @@
...
+1). E2E examples are directly located under helm-charts/ directory
+2). Charts for components are located under helm-charts/common directory.
+
+We have Xeon servers with Gaudi accelerator card, so basicly we'll cover running workloads for cpu only and with Gaudi accelerated.For other configurations/feature coverage, it's case by case.
```
eero-t, line 14: There's typo, but IMHO this can be simplified to:
```suggestion
We have Xeon servers with Gaudi accelerator card, which covers CPU and Gaudi accelerated workloads. For coverage of other configurations/features, it's case by case.
```

helm-charts/CI.md
```diff
@@ -0,0 +1,26 @@
...
+
+We'll run CI test when detected values file beginning with "ci-" and ending with "values.yaml"(`ci-*values.yaml`).
```
eero-t, line 16: Simpler e.g. as:
```suggestion
CI tests values files which names are listed in `ci-*values.yaml` files.
```

helm-charts/CI.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Add new test case
+
+Add a new values file following the naming rules: `ci-*values.yaml`.
+If the values file for CI is exactly the same with another one, you can just use a symbol link.
```
eero-t, line 21:
```suggestion
If the values file for CI is identical to another one, a symbolic link can be used.
```

eero-t (commented):
Few suggestions for the CI document.

yongfengdu:
Check Path error is due to CI script error, I've double checked the link is valid.
Guardrails test failure is not related to this change, and need further debug.
Since this PR will trigger all CI tests for helm charts and consume quite a lot resource and will block other changes, let's merge this first and I'll provide a follow up fix for retriever-usvc and prompt-usvc and disable guardrails CI.

lianhao (approved):

lianhao:
> Check Path error is due to CI script error, I've double checked the link is valid. Guardrails test failure is not related to this change, and need further debug. Since this PR will trigger all CI tests for helm charts and consume quite a lot resource and will block other changes, let's merge this first and I'll provide a follow up fix for retriever-usvc and prompt-usvc and disable guardrails CI.

fast approve this due to above reasons

daisy-ycguo (approved):

PR #509, opened 2024-10-29
CodeGen on OpenShift

polszewska:
That part allows to deploy CodeGen on Red Hat OpenShift. It contains two options: with Red Hat OpenShift AI and without. Both variants were prepared to run on Xeon or using Gaudi accelerators.

### DIFF
--- a/helm-charts/codegen-openshift-rhoai/Chart.yaml
+++ b/helm-charts/codegen-openshift-rhoai/Chart.yaml
@@ -1,3 +1,6 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
 apiVersion: v2
 name: codegen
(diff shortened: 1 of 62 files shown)


lianhao:
Just wondering whether we can combine the ordinary helm charts with these 2 OpenShift variants, with some switches to control whether it's deployed on normal k8s, OpenShift and/or OpenShiftAI. Otherwise, there would be 3 versions of llm-uservice, tgi, codegen, etc, which is very confusing to end users if we publish the helm charts [in the chart repo](https://opea-project.github.io/GenAIInfra/)(which we've already did, and is planning to improve it).

polszewska:
Technically it is possible, but our team was terminated during the current CPMs, so we won’t continue working on OPEA enablement on OpenShift. If you’re interested please feel free to use these charts to adapt them to your needs.

PR #523, opened 2024-11-04
Rename Helm HPA option + fixes to Helm READMEs

eero-t:
## Description

Rename `horizontalPodAutoscaler` option as `autoscaling`, to align it with the name used in Nvidia Helm charts:
https://github.com/NVIDIA/nim-deploy/blob/main/helm/nim-llm/README.md#autoscaling-parameters

And fix how `bool` options are shown in Helm READMEs.

## Issues

`n/a`.

## Type of change

Fixes README typos, changes (very rarely used) Helm option name:
- [x] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

`n/a`.

## Tests

Change is automated (+ some README editing to align table columns):
`git grep -l horizontalPodAutoscaler | xargs sed -i s/horizontalPodAutoscaler/autoscaling/g`

And from the diff it's clearly doing the right thing so there should be no need for testing.  If things worked earlier, they still work.


lianhao (approved):

helm-charts/HPA.md
```diff
@@ -15,7 +15,7 @@
 
 ## Introduction
 
-`horizontalPodAutoscaler` option enables HPA scaling for relevant service components:
+`autoscaling` option enables HPA scaling for relevant service components:
 https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
```
poussa, line 19: I would like to use  just `hpa` instead of `autoscaling`, and then open that as Horizontal Pod Autoscaling in the text. `autoscaling` is to generic and can mean horizontal or vertical pod, or cluster (node) autoscaling.

eero-t: Is cluster node autoscaling something that Helm charts would control (install rules for), and something that could be added to OPEA Helm charts later on?    And if yes, why that would not be also be under `autoscaling` option:
* `autoscaling.scaleNodes: true`
* `autoscaling.maxNodes: 99`
?

poussa: `autoscaling.hpa` is good. We can extend the `autoscaling` in the future then, if needed.

poussa: yes, that is the case. My point is that `autoscaling` is too generic. If this is about HPA autoscaling, lets have the `hpa` in the name. So, `hpa` or `autoscaling.hpa` are both fine.

eero-t: > `autoscaling.hpa` is good. We can extend the autoscaling in the future then, if needed.

As cluster scaling is not really an application install setting, there should be no confusion with that.

Are there also other relevant ways of scaling k8s _applications_ than HPA?

(`.enabled` -> `.hpa` change is OK to me, but I'm not yet seeing compelling reason to differ from NIM Helm charts.)

poussa: Yeah, the NIM point is valid. Let's keep it similar.

poussa (approved):

PR #587, opened 2024-11-19
Use default kubernetes imagePullPolicy

yongfengdu:
For latest tag, it should use Always.
For tags other than latest, it will use IfNotPresent.
https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

## Description

## Issues

#597 

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/agentqna/Chart.yaml
+++ b/helm-charts/agentqna/Chart.yaml
@@ -7,37 +7,37 @@ description: The Helm chart to deploy AgentQnA
 type: application
 dependencies:
   - name: agent
-    version: 1.0.0
+    version: 0-latest
     alias: worker
     repository: "file://../common/agent"
   - name: agent
-    version: 1.0.0
+    version: 0-latest
     alias: supervisor
     repository: "file://../common/agent"
   - name: tgi
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/tgi"
     condition: tgi.enabled
   - name: tei
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/tei"
   - name: embedding-usvc
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/embedding-usvc"
   - name: teirerank
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/teirerank"
   - name: reranking-usvc
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/reranking-usvc"
   - name: redis-vector-db
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/redis-vector-db"
   - name: retriever-usvc
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/retriever-usvc"
   - name: data-prep
-    version: 1.0.0
+    version: 0-latest
     repository: "file://../common/data-prep"
-version: 1.0.0
+version: 0-latest
 appVersion: "v1.0"
--- a/.github/workflows/manual-freeze-tag.yaml
+++ b/.github/workflows/manual-freeze-tag.yaml
@@ -7,17 +7,17 @@ on:
   workflow_dispatch:
     inputs:
       oldversion:
-        default: "1.1-dev"
+        default: "0-latest"
         description: "Old helm version to be replaced"
         required: true
         type: string
       newversion:
-        default: "1.1.0"
+        default: "1.2.0"
         description: "New helm version to replace"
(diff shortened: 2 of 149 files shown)


eero-t:
> Target for 1.2-dev. We won't fix this for 1.1 release because IfNotPresent is proper policy for released version.

That *is* the default when policy is not specified, and tag is something else than "latest".

lianhao:
agreed with @mkbhanda , I think we can make it looks nicer by:

in values.yaml, leave the image.pullPolicy as a comment with a little bit explanation, which gives the user some hints that it's configurable, i.e.
```
image:
  repostitory: opea/abc
  tag: "latest"
  # Uncomment the following line to set image desired image pull policy if needed, as one of Always, IfNotPresent, Never
  # pullPolicy: ""
```

In the corresponding deployment.yaml, change the image policy settings as optional:
```
   ... ...
   {{- if .Values.image.pullPolicy }}
   imagePullPolicy: {{ .Values.image.pullPolicy }}
   {{- end }}
```

helm-charts/agentqna/Chart.yaml
```diff
@@ -7,37 +7,37 @@ description: The Helm chart to deploy AgentQnA
 type: application
 dependencies:
   - name: agent
-    version: 1.0.0
+    version: 0-latest
```
eero-t, line 10: Why `0-latest` instead of `1.1.0`?

eero-t: Using 0 as version number seems rather misleading.  I think something like `1.2-dev`, `1.2-rc` or `1.1.0-latest` would be better.

eero-t: Best would be if "NeuroBot" would do such update to `main` automatically, right after it has updated versions for the release branch...

yongfengdu: There should be just one development version in the packages (https://github.com/orgs/opea-project/packages/container/package/charts%2Fchatqna). If we use 1.2-dev, there would be more versions like 1.3-dev remains, and it's hard to cleanup. Actually 1.2-dev was the name I first used, but think 0-latest from @poussa is a better one. I'm open to other good proposals.

There should be no such update on main branch after this one merged, so NeuroBot would not be necessary.

yongfengdu:
The helm lint fail is due to "helm dep build", refer to this:
https://github.com/helm/helm/issues/8036
Will need to add the repo first as @eero-t mentioned.

yongfengdu:
Checked the GMC CI chatqna.log, the output file is good with "[DONE]" ending. That's more like an issue from client side, or the "router" sevice of GMC.
> * GMC test:
>   
>   * `Response check failed, please check the logs in artifacts`
>   * => CI checks response before it has fully completed?

Gaudi device conflict, rerun should solve this.
> * ChatQnA, TGI & vLLM tests:
>   
>   * `Shard 0 failed to start`
>   * `RuntimeError: synStatus=8 [Device not found] Device acquire failed`
>   * => CI Gaudi k8s device plugin error; pod resource request is satisfied, but device is missing?

eero-t:
Ok, the other CI issues are now fixed. The 2 remaining ones (audioqna + speecht5 on Gaudi) are both due to CI using stale `opea/speecht5-gaudi:latest` image with this issue:
```
 Traceback (most recent call last):
  File "/home/user/comps/tts/speecht5/dependency/speecht5_server.py", line 15, in <module>
    from comps import CustomLogger
  File "/home/user/comps/__init__.py", line 7, in <module>
    from comps.cores.proto.docarray import (
  File "/home/user/comps/cores/proto/docarray.py", line 10, in <module>
    from pydantic import Field, conint, conlist, field_validator
ImportError: cannot import name 'field_validator' from 'pydantic' (/usr/local/lib/python3.10/dist-packages/pydantic/__init__.cpython-310-x86_64-linux-gnu.so)
```

helm-charts/agentqna/Chart.yaml
eero-t: > Actually 1.2-dev was the name I first used, but think 0-latest from @poussa is a better one.

@poussa Why `0-latest` instead of `latest`?

lianhao (approved):
The speecht5-gaudi issue is reported at  opea-project/GenAIComps#948, which is not related to this change.

lianhao:
We'll let the PR merge in first even with issue opea-project/GenAIComps#948 to avoid huge CI tests cases involved in this PR

PR #618, opened 2024-11-28
helm: Add service account support - part2

### LINKED ISSUE #532: [Feature][helm-charts] Expose service account
In order to make OPEA deployments easier and more secure the service account (SA) needs to be exposed to deployments. Especially to those accessing the model data (e.g. tgi). In CSP environments (e.g. Google cloud), the model data is often stored in a cloud service (e.g., Google Cloud Storage (gcs)). Pod access to these services are managed via SAs. Special SAs are created and granted access. We don't want to use `default` SA since it gives all pods access, which is unnecessary. 

Currently, there is no way to set the SA for any deployments. 

In general, it would be good if all the helm charts settings are based on what comes out from `helm create`. The settings are there for a reason.

lianhao:
## Description

Add service account support in the following helm charts:

- guardrails-usvc
- llm-uservice
- lvm-uservice
- mongodb
- prompt-usvc
- redis-vector-db
- reranking-usvc
- retriever-usvc
- tts
- ui
- web-retriever

## Issues

Fixes #532.

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/common/guardrails-usvc/templates/deployment.yaml
+++ b/helm-charts/common/guardrails-usvc/templates/deployment.yaml
@@ -25,6 +25,7 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      serviceAccountName: {{ include "guardrails-usvc.serviceAccountName" . }}
       securityContext:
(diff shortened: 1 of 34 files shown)


lianhao:
close due to #624

PR #624, opened 2024-12-04
helm chart: Add service account support

### LINKED ISSUE #532: [Feature][helm-charts] Expose service account

lianhao:
## Description

1. Add service account support disabled by default for all common charts, enabled by default for all e2e charts.

2. Add support of sharing the same service account by setting `global.sharedSAName`, disabled by default for all common charts, enabled by default for all e2e charts.

## Issues

Fixes #532.

## Type of change

## Dependencies

n/a.

## Tests

### DIFF
--- a/helm-charts/agentqna/templates/_helpers.tpl
+++ b/helm-charts/agentqna/templates/_helpers.tpl
@@ -54,7 +54,9 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Create the name of the service account to use
 */}}
 {{- define "agentqna.serviceAccountName" -}}
-{{- if .Values.serviceAccount.create }}
+{{- if .Values.global.sharedSAName }}
+{{- .Values.global.sharedSAName }}
+{{- else if .Values.serviceAccount.create }}
 {{- default (include "agentqna.fullname" .) .Values.serviceAccount.name }}
 {{- else }}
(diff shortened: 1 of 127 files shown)


lianhao:
The failure is not related to this PR, please see issue opea-project/GenAIComps#969

poussa (approved):

yongfengdu (approved):

lianhao:
All CI jobs have passed except for the known issue https://github.com/opea-project/GenAIComps/issues/969 which is not related to this PR.

I'll merge this PR first to avoid rerun dozens of CI jobs again.

eero-t:
@poussa Service account name is not prefixed with Helm chart release name, like is the case with all the other created k8s objects.  Is that a problem?

PR #636, opened 2024-12-11
Use unified nginx helm chart

### LINKED ISSUE #635: [Feature] Use unified nginx chart
### Priority

P3-Medium

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Running nodes

Single Node

### Description

Current we don't directly use opea/nginx image for reverse proxy, which could make us difficult to follow the changes in opea/nginx microservice. Also, it created duplications in the existing e2e helm charts.

We should use unified helm chart for opea/nginx to minimize maintenance effort.

lianhao:
## Description

- Add helm chart support for OPEA nginx

- Use unified nginx chart in E2E.

## Issues

Fixes #635.

## Type of change

## Dependencies

## Tests


helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+# Enabling HPA will:
+# - Ignore above replicaCount, as it will be controlled by HPA
+# - Add example HPA scaling rules with thresholds suitable for Xeon deployments
+# - Require custom metrics ConfigMap available in the main application chart
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 4
+  # targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
eero-t, line 87: Did you test autoscaling?   I would think HPA deployment to fail when there's no threshold specified for scaling...

helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+# Additional volumes on the output Deployment definition.
+volumes: []
+#  - name: tmp
+#    emptyDir: {}
+
+# Additional volumeMounts on the output Deployment definition.
+volumeMounts: []
+#  - name: tmp
+#    mountPath: /tmp
```
eero-t, line 97: What volumes NGINX is expected to need?

(IMHO values files should not have items that are never going to be used.)

helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+
+# Name of the ConfigMap to be used as the environment variables
+# See https://github.com/opea-project/GenAIComps/blob/main/comps/nginx/nginx.conf.template for details
+cmName: ""
```
eero-t, line 107: Applications' hard-code NGINX `configMap` name to  `{{ .Release.Name }}-nginx-config`.

So I do not see a point in having this configurable.  Just hard-coded it in NGINX `Deployment` to a name applications will use, like all the other configMaps for common components do.

helm-charts/common/nginx/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "nginx.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          {{- if .Values.cmName }}
```
eero-t, line 38: Does NGINX do anything useful for the applications without properly configured env vars in `configMap`?

If not, there's no point in making this optional.

helm-charts/common/nginx/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Helm chart for deploying OPEA nginx service.
+
+## Using nginx chart
+
+This chart is not meant to be used standalone. It should be used with OPEA E2E helm charts as the reverse proxy for UI.
+
+1. In the OPEA E2E chart, define a ConfigMap containing the environment variables as defined in [OPEA nginx microservice](https://github.com/opea-project/GenAIComps/blob/main/comps/nginx/nginx.conf.template).
+
+2. Pass the name of the ConfigMap to helm configuration `cmName`.
```
eero-t, line 11: With the redundant cm name configuration removed:
```suggestion
E2E charts for OPEA applications provide ConfigMap with the required environment variables to configure [OPEA nginx microservice](https://github.com/opea-project/GenAIComps/blob/main/comps/nginx/nginx.conf.template).
```

helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+# This section is for setting up autoscaling more information can be found here: https://kubernetes.io/docs/concepts/workloads/autoscaling/
+# Enabling HPA will:
+# - Ignore above replicaCount, as it will be controlled by HPA
+# - Add example HPA scaling rules with thresholds suitable for Xeon deployments
+# - Require custom metrics ConfigMap available in the main application chart
```
eero-t, line 81: HPA does not ignore replicaCount, value=1 is ignored regardless of HPA (as it's k8s default), there's no acceleration alternative for this, and you used k8s base metrics, instead of app/custom metrics:
```suggestion
# Note: do not use (above) "replicaCount" with HPA (Chart ignores value=1 as it's k8s default)
```

helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,113 @@
...
+service:
+  # This sets the service type more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
+  type: ClusterIP
+  # This sets the ports more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
+  port: 80
+
+resources: {}
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
```
eero-t, line 68: Please do not copy this bogus reasoning.  Minikube[1]  surely cannot be a reason for breaking OPEA applications for real setups?

With reasonable resource declarations pod gets assigned to a node suitable for it, or if none is available, it remains  in `Pending` state, **instead** of using constantly CPU in crash loop and causing (potentially) other OPEA pods (with no resource allocs i.e. lowest priority ones) being evicted from the node.

I thnk that for pods which resource usage does not depend on used arguments (like is the case with inferencing services), requesting fixed amount of resources should be reasonable, and it's important for HPA.

[1] For something like that, better would be having separately doc for "Running OPEA app in resource constrained test setup" detailing how to specify smaller inferencing models and remove resource requests for pods in hopes that components won't end up in crash loops.

eero-t (changes requested):

lianhao:
The faqgen failure is due to the known issue of opea-project/GenAIComps#969

helm-charts/common/nginx/values.yaml
lianhao: The reason I leave this configurable is to make CI happy. Otherwise, a hardcoded configmap would prevent the nginx's own CI test failure due to unable to schedule pod, because of this configmap is defined in the application e2e chart, not in the nginx chart itself.

helm-charts/common/nginx/values.yaml
lianhao: done

helm-charts/common/nginx/values.yaml
lianhao: will change

helm-charts/common/nginx/templates/deployment.yaml
lianhao: without proper configmap, this nginx won't be useful at all. Here I make it optional to please the current CI.

helm-charts/common/nginx/values.yaml
lianhao: done

helm-charts/common/nginx/values.yaml
lianhao: This paragraph is generated by 'helm create'. You're right, we should clean it up.

lianhao: done

lianhao:
we need to wait until [langserve 0.3.1 is available in pypi](https://pypi.org/project/langserve/#history) to unblock the faqgen CI failure

helm-charts/common/nginx/templates/deployment.yaml
lianhao: done, make this optional to please CI

helm-charts/common/nginx/values.yaml
lianhao: done

lianhao:
gaudi CI failure: 
`huggingface_hub.utils._errors.HfHubHTTPError: 429 Client Error: Too Many Requests for url: https://huggingface.co/api/whoami-v2`

yongfengdu (approved):

helm-charts/common/nginx/values.yaml
eero-t: Just to verify... You tested that Helm install with HPA enabled, works even when these values are commented out?

helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+  #   memory: 128Mi
+
+# This section is for setting up autoscaling more information can be found here: https://kubernetes.io/docs/concepts/workloads/autoscaling/
+# Note: do not use (above) "replicaCount" with HPA (Chart ignores value=1 as it's k8s default)
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 4
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
eero-t, line 83: IMHO these variable names could be a bit shorter.
```suggestion
  targetCPUPercentage: 80
  # targetMemoryPercentage: 80
```

helm-charts/common/nginx/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+resources: {}
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
```
eero-t, line 74: I've never used these (only application specific custom metrics), but according to docs: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#how-does-a-horizontalpodautoscaler-work

CPU/mem resource requests need to be specified for HPA to work on relative CPU/mem targets:

> For per-pod resource metrics (like CPU), the controller fetches the metrics from the resource metrics API for each Pod targeted by the HorizontalPodAutoscaler. Then, if a target utilization value is set, the controller calculates the utilization value as a percentage of the equivalent [resource request](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits) on the containers in each Pod.
> ...
> Please note that if some of the Pod's containers do not have the relevant resource request set, CPU utilization for the Pod will not be defined and the autoscaler will not take any action for that metric.

eero-t: So I think that:
* `requests` part should be enabled, after verifying that current values match specified nginx version resource usage
* Comment for `requests` values should state that "normal" nginx resource usage should be checked and values updated accordingly, before enabling autoscaling
* Comment for limits needs to state that they should be enabled only after checking how much nginx needs **when it's stressed**, and using those values _+ some extra for growth_ (increased buffers, possible leakage etc)

helm-charts/common/nginx/values.yaml
lianhao: yes

helm-charts/common/nginx/values.yaml
lianhao: Can we do this in another PR along with issue #643 ?

lianhao:
The docsum failure will be fixed by incoming PRs for issue #658

helm-charts/common/nginx/values.yaml
lianhao: These names are from the default helm starter template

helm-charts/common/nginx/values.yaml
eero-t: > Can we do this in another PR along with issue https://github.com/opea-project/GenAIInfra/issues/643 ?

OK, but HPA won't work for this component before that...

helm-charts/common/nginx/values.yaml
eero-t: Ok

eero-t (approved):
Approved on assumption that resource requests are defined at least for HPA related pod(s) in successive PR(s).

Ruoyu-y (approved):
LGTM

PR #655, opened 2024-12-20
Add monitoring for rest of ChatQnA + DocSum components

eero-t:
## Description

ChatQnA + DocSum had few components which provide metrics endpoint, but did not have Prometheus `serviceMonitor`s for (optional) collection of those metrics.  This adds them.

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

Manual testing with `latest` images.

## Next steps

These components use `MicroService` class, so they provide metrics both from that and underlying HTTP service.  

However, because they don't process any tokens with `MicroService`, its metrics are always zero.  I'll try to come up with some way to skip exporting them from `MicroService` in "GenAIComps".

### DIFF
--- a/helm-charts/common/data-prep/README.md
+++ b/helm-charts/common/data-prep/README.md
@@ -46,12 +46,13 @@ curl http://localhost:6007/v1/dataprep  \
 
 ## Values
 
-| Key                    | Type   | Default                 | Description |
-| ---------------------- | ------ | ----------------------- | ----------- |
-| image.repository       | string | `"opea/dataprep-redis"` |             |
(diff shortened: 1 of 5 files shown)


lianhao (approved):

yongfengdu (approved):

eero-t:
> These components use `MicroService` class, so they provide metrics both from that and underlying HTTP service.
> 
> However, because they don't process any tokens with `MicroService`, its metrics are always zero. I'll try to come up with some way to skip exporting them from `MicroService` in "GenAIComps".

=> https://github.com/opea-project/GenAIComps/pull/1092

Issue #643, opened 2024-12-13
[Feature] Helm charts needs to have minimal resource requirements specified in default values for all containers


kad:
### Priority

Undecided

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Running nodes

Multiple Nodes

### Description

Currently, all the containers in helm charts by default are missing resource requests/limits. This makes the Pod QoS in Kubernetes to be BestEffort, which is least priority workload on the node, leading to instability of performance.

Default values for all containers should be requesting some minimal amount of CPU and memory resources.
We don't need to specify `limits` for now, but minimal `requests` are mandatory to make containers into Burstable Pod QoS.

Minimal values can be easily collected from any of the benchmark runs, during "steady" state of execution. Values not needed to be precise, but some minimal viable numbers should be added.

lianhao:
The minimum resource requests will be very different for different models for those components directly using LLM models. For other components which doesn't directly involves models, that could be done.

kad:
Even for different LLM models, we can have some minimal requirements: e.g. 1 vcpu, 100Mb of RAM.
For model which we consider default in helm chart, we can add more realistic minimal requirements.

eero-t:
@kad There's some discussion about how to best handle the problem with (model) limits here: https://github.com/opea-project/GenAIInfra/pull/431

kad:
> @kad There's some discussion about how to best handle the problem with (model) limits here: #431

that discussion in https://github.com/opea-project/GenAIInfra/pull/431 seems to be stalled, but yes, my point is along the lines that is mentioned in some of the comments:
- we have worst Pod QoS now without specifying anything: Best Efforts. We need to get all containers to Burstable QoS.
- limits are really model specific and might be trieckier to manage. However, for this issue it is not needed currently. We need only requests to be spcified.
- Too big requests can be a problem, but we need to start with some minimal viable values, based on current observations.

yongfengdu:
Do we need to set the minimal resource request for all helm charts, or just the compute intensive one like TEI/TGI/vLLM?

eero-t:
> Do we need to set the minimal resource request for all helm charts, or just the compute intensive one like TEI/TGI/vLLM?

All should have them.  Application is not much use if backend inferencing is idling because services before it are throttled, or even being evicted.

Suitable (minimum) requests should be much easier to specify for those other services, as their resource usage does not depend on which model user has set in Helm values.

yongfengdu:
[...]

PR #662, opened 2024-12-31
Add minimal resource requests for tgi

yongfengdu:
## Description

This is the first PR for #643 , the plan is to get tgi change reviewed and merged, then doing the same for other charts.(To avoid too many CI testing)
Tried several requests settings(With small values) and it doesn't have obvious impact on the resource usage.
So set this should be good enough.
cpu: 100m
memory: 128Mi

It's performance tuning's task to set more reasonable resource limit/requests according to different scenarios.

## Issues

#643 

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/common/tgi/values.yaml
+++ b/helm-charts/common/tgi/values.yaml
@@ -69,17 +69,13 @@ securityContext:
 service:
   type: ClusterIP
 
-resources: {}
-  # We usually recommend not to specify default resources and to leave this as a conscious
-  # choice for the user. This also increases chances charts run on environments with little
-  # resources, such as Minikube. If you do want to specify resources, uncomment the following
-  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+resources:
   # limits:
   #   cpu: 100m
   #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  requests:
+    cpu: 100m
+    memory: 128Mi
 
 # Use TCP probe instead of HTTP due to bug #483
 # https://github.com/opea-project/GenAIExamples/issues/483


helm-charts/common/tgi/values.yaml
```diff
@@ -69,17 +69,13 @@ securityContext:
...
+resources:
   # limits:
   #   cpu: 100m
   #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  requests:
+    cpu: 100m
+    memory: 128Mi
```
lianhao, line 78: Will these numbers are too small? Have you tested it with a minimal model?

yongfengdu: The requests are just minimal requests used by scheduler, the actual usage is not affected by this setting.
kubectl describe pod xxx shows that it's Burstable.
QoS Class:                   Burstable

The actual resources are not affected(The CPU cores could be 6000m CPUs when busy, but is just 2m at idle):
NAME                    CPU(cores)   MEMORY(bytes)
tgi-57c674bfb8-k2h4m    1002m        15773Mi
tgi1-55558f6c9-kmqr9    1002m        15779Mi
tgi2-7854bf5569-qm5p2   2m           16232Mi

eero-t: For the priority class improvement, request can be for service resource usage when it's idling.  This is OK for start.

However, a service that has resources just for idling (when node is otherwise stressed), is not a very useful one. 

For better resource requests, values should correspond to active service usage.  Problem with that is lack of service SLA definitions for the OPEA services, what kind of load they're supposed to handle, on what HW, and with what performance?

Once the load & HW parts are defined, active service usage can be tested, and corresponding service resource usage determined for the resource requests.

(In TGI case where resource usage is dictated by used model, and options related to it, SLA should define also those.)

lianhao (approved):
Let's have this as a starter and put more meaningful values once the opea SLA is defined.

poussa (approved):

Issue #596, opened 2024-11-21
[Feature] Helm Charts for Txt2Img and SearchQnA.


joshuayao:

yongfengdu:
Any priority list for the GenAIExamples? If we don't have enough resource to add all examples, we need priority list to decide which to go first.
Besides adding more Examples, I think the most important thing to do is fix issues and enhance the high priority Examples/Components to make them ready for production.

joshuayao:
Hi @yongfengdu. According to the roadmap, Helm Charts will be the default deployment method in v1.2. To ensure consistency across examples, it would be ideal to update all examples to support Helm Charts. I completely understand your concerns. For any changes to the feature scope of the roadmap, please reach out to Ke and Malini.

joshuayao:
Hi @yongfengdu, is your team working on this?

yongfengdu:
As I know the GenAIExamples is under refactoring for 1.2, and helm-chart will need to change accordingly.
It's not wise to add more examples if they will be merged soon.
I'd propose to postpone the "more examples support" to 1.3 release, since it would be a waste of engineering resource if we're doing sth like this:
https://github.com/opea-project/GenAIInfra/pull/590  - Add more microservices for docsum
https://github.com/opea-project/GenAIInfra/pull/659  - Reduce microservices in docsum

The later one removed >80% of the code for previous PR.

yongfengdu:
Walked through the GenAIExamples' current list, here is the summary for gaps:
* Already supported, will follow up with update
AgentQnA, AudioQnA, ChatQnA, CodeGen, CodeTrans, DocSum, FaqGen, VisualQnA

* Not yet supported, but is planning to do (5 Examples).
AvatarChatBot - AudioQnA+wav2lip+animation, 2 more microservices required.
DBQnA - Use LLM to Generate SQL. CPU only. More components required postpres, text2sql-service
GraphRAG - Variants of ChatQnA, neo4j graph database required, Gaudi only
SearchQnA - Use webretriever. Web-retriever supported, CPU only
Text2Image - No compose file, Components only
 
* Not support. no plan to support.
DocIndexRetriever - Not an E2E example, part of AgentQnA
EdgeCraftRag	- ARC GPU only, all functions are implemented in edgecraftrag-server, not using GenAIComps. Less likely to run on k8s environmetn, defer the support or more discuss/requirement needed.
InstructionTuning/RerankingTuning - Not E2E example, No compose file
[...]

PR #725, opened 2025-01-18
Add Text2Image e2e helm chart

yongfengdu:
There is no mega service for text2image, just add an options to the components.

## Description

## Issues

Part of #596 

## Type of change

## Dependencies

## Tests


lianhao:
I'm mostly fine to this, just wondering should we move this to e2e layout, just like others?

yongfengdu:
> I'm mostly fine to this, just wondering should we move this to e2e layout, just like others?

Because the name in GenAIExamples and GenAIComps are the same, in this case we'll have to choose a different name for the Example, I'll use txt2img as the example name.

lianhao (approved):

PR #774, opened 2025-02-05
Helm chart for Ollama

jonminkin97:
## Description

Add Helm chart to deploy Ollama as a model server.

## Issues

n/a

## Type of change

## Dependencies

- [Ollama](https://github.com/ollama/ollama)

## Tests

- Manual deployment of Helm chart, verifying installation and inference
- Helm test pod to automatically test the Helm release using `helm test [RELEASE]`


helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+  httpGet:
+    path: /api/version
+    port: http
+readinessProbe:
+  httpGet:
+    path: /api/version
+    port: http
+
+
+OLLAMA_MODEL: "llama3.2:1b"
```
lianhao, line 89: I would suggest to use `LLM_MODEL_ID` here, to use the same config name as vllm/tgi, because ollama is another inference engine like vllm/tgi in OPEA.

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
+# This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/
+service:
+  # This sets the service type more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
+  type: ClusterIP
+
```
lianhao, line 60: remove unused blank line pls

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+  port: 80
+
```
lianhao, line 62: remove unused blank line pls

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+# For more information checkout: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+podAnnotations: {}
+# This is for setting Kubernetes Labels to a Pod.
+# For more information checkout: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+podLabels: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext: {}
```
lianhao, line 48: Please consider adding some minimal securityContext settings if possible.

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+replicaCount: 1
+
+# This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
+image:
+  repository: ollama/ollama
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "0.5.7"
+
+  # Uncomment the following line to set desired image pull policy if needed, as one of Always, IfNotPresent, Never.
+  pullPolicy: Always
```
lianhao, line 18: Please comment this line, and let the k8s default kicks in. see https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/mm-embedding/values.yaml#L26

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+readinessProbe:
+  httpGet:
+    path: /api/version
+    port: http
+
+
+OLLAMA_MODEL: "llama3.2:1b"
+
+global:
+  http_proxy: ""
```
lianhao, line 92: Please add `sharedSAName` support, just like https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/lvm-serve/values.yaml#L179

helm-charts/common/ollama-service/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+*/}}
+{{- define "ollama-service.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "ollama-service.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "ollama-service.serviceAccountName" -}}
```
lianhao, line 56: Please follow the convention in https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/lvm-serve/templates/_helpers.tpl#L56 to add the global sharedSAName support, so all the components in a e2e helm charts can share the same service account name

helm-charts/common/ollama-service/templates/deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+            - mountPath: /root/.ollama/models
+              name: model-volume
+            {{- with .Values.volumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+      volumes:
+        - name: model-volume
```
lianhao, line 73: Please support `modelUsePVC` and `modelUseHostPath`, just follow https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/lvm-serve/templates/_helpers.tpl#L56

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+# This is to setup the liveness and readiness probes more information can be found here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+livenessProbe:
+  httpGet:
+    path: /api/version
+    port: http
```
lianhao, line 82: Maybe we could delete this livenessProbe if we don't see pod got stuck during processing requests.

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+readinessProbe:
```
lianhao, line 83: Maybe we should consider adding some startupProbe in case the model warming up takes quite some time?

helm-charts/common/ollama-service/templates/deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
```
lianhao, line 88: Please add the topologySpreadConstraints support which is used by benchmarking. Just like https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/lvm-serve/templates/deployment.yaml#L159

lianhao (changes requested):
Thanks very much for contributing this. Besides the embedded comment, there is something missing in the helm template(e.g. HPA, service monitor). I would suggest you check https://github.com/opea-project/GenAIInfra/tree/main/helm-charts/common/lvm-serve to add the missing gap. Thanks!

helm-charts/common/ollama-service/values.yaml
jonminkin97: Ollama loads models into memory using lazy loading which is triggered by an initial inference. I added a startup probe to verify that the model is downloaded.

jonminkin97:
@lianhao Thank you for the feedback. I've made the requested changes, and they are now pending your approval. Let me know if there are any other enhancements you would like to see.

lianhao:
@jonminkin97 If you could address the 2 following things I think this PR is ready to go:

- Add a cpu-values.yaml file which is used to trigger the CI, just like what [lvm-serve](https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/lvm-serve/cpu-values.yaml) does.

- Could you name this chart ollama instead of ollama-service? One of our 1.3 task is to make sure the helm chart name is consistent with the corresponding names in GenAIComps.

helm-charts/common/ollama-service/templates/deployment.yaml
```diff
@@ -0,0 +1,130 @@
...
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command: ['bash', '-c']
+          args:
+            - ollama serve & sleep 10 && ollama pull ${LLM_MODEL_ID}
+          volumeMounts:
+            # Running as non-root sets $HOME to /
+            - mountPath: /.ollama
+              name: model-volume
+
```
lianhao, line 57: It'd be better to remove this empty line

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+  # fsGroup: 2000
+
+securityContext:
+  capabilities:
+    drop:
+    - ALL
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 1001
+  runAsGroup: 1001
```
lianhao, line 56: we can leave runAsGroup unset, since it's not required in [k8s pod security standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/)

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext:
+  capabilities:
+    drop:
+    - ALL
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 1001
```
lianhao, line 55: Any reason that we use uid 1001, instead of conventional 1000?

helm-charts/common/ollama-service/templates/deployment.yaml
```diff
@@ -0,0 +1,130 @@
...
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+            capabilities:
+              drop:
+                - ALL
+              add:
+                - DAC_OVERRIDE
+                # To be able to make data model directory group writable for
+                # previously downloaded model by old versions of helm chart
+                - FOWNER
```
lianhao, line 47: Do we need FOWNER capability here? It's used for the command of 'chmod' in other inference engine case.

lianhao (commented):
2 more things beside the embedding comments:

- Add a cpu-values.yaml file which is used to trigger the CI, just like what [lvm-serve](https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/lvm-serve/cpu-values.yaml) does. These kind of xxx-values contains additional configuration for CI to pass(could be empty if there no additional requirements at all).

- Could you name this chart ollama instead of ollama-service? One of our 1.3 task is to make sure the helm chart name is consistent with the corresponding names in GenAIComps.

helm-charts/common/ollama-service/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+# - stop accepting new requests
+# - handle all of its buffered requests
+# - terminate after tthose have been processed
+# See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination.
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 4
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
eero-t, line 102: If this is only downloading model data, why it would use so much CPU that its replica count would need to increased?

And wouldn't it in that case make more sense to request it to be scheduled somewhere where there's more that 1 CPU core free?  I.e. set resources to appropriately to expected CPU utilization?

Also, I don't think `evenly_distributed` setting makes sense here.  Setting that would schedule this service to same node where the service using the downloaded model is, would seem to make more sense...

helm-charts/common/ollama-service/templates/deployment.yaml
```diff
@@ -0,0 +1,130 @@
...
+      {{- end }}
+      {{- if .Values.evenly_distributed }}
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: kubernetes.io/hostname
+          whenUnsatisfiable: ScheduleAnyway
+          labelSelector:
+            matchLabels:
+              {{- include "ollama-service.selectorLabels" . | nindent 14 }}
+      {{- end }}
```
eero-t, line 130: I think setting that would schedule this service to same node where the service using the downloaded model is, would make more sense, than just randomly spreading download service around the cluster.

helm-charts/common/ollama-service/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,18 @@
...
+  labels:
+    release: {{ .Values.global.prometheusRelease }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "ollama-service.selectorLabels" . | nindent 6 }}
+  endpoints:
+  - port: ollama-service
+    interval: 5s
+{{- end }}
```
eero-t, line 18: According to this ticket, Ollama does not provide Prometheus `/metrics` endpoint yet, so this is redundant:
https://github.com/ollama/ollama/issues/3144

yongfengdu (approved):

PR #782, opened 2025-02-06
Enable passing HF_TOKEN to web-retriever

yongfengdu:
## Description

## Issues

https://github.com/opea-project/GenAIComps/issues/1264

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/common/web-retriever/templates/configmap.yaml
+++ b/helm-charts/common/web-retriever/templates/configmap.yaml
@@ -25,3 +25,6 @@ data:
   {{- end }}
   HF_HOME: "/tmp/.cache/huggingface"
   LOGFLAG: {{ .Values.LOGFLAG | quote }}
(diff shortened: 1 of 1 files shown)


eero-t (approved):
Looks fine, but I'm curious what this component should retrieve from HF?

yongfengdu:
> Looks fine, but I'm curious what this component should retrieve from HF?
You can see the url from the pasted pic of https://github.com/opea-project/GenAIComps/issues/1264
To be honest, it doesn't make sense to me why it would access that URL, but that would be another story to fix that.

lianhao (changes requested):
Can we double check if the HF_TOKEN is still required? If so, we should have it in both README and dafault values.yaml

yongfengdu:
There is no reason web-retriever should access huggingface with HF_TOKEN, I also tried with the latest image, and it works fine without HFT_TOKEN. So we'll not merge this.
Will keep this PR as draft for a while just in case we'll need it later.

Issue #772, opened 2025-02-05
[ci-auto] Update vllm & tgi to support DeepSeek models


CICD-at-OPEA:
## GenAIExample ChatQnA compose.yaml got changed

Below files are changed in [this commit](https://github.com/opea-project/GenAIExamples/commit/9adf7a6af003e505b9acf0c1b1871e86edc4967d)

- ChatQnA/docker_compose/intel/hpu/gaudi/compose.yaml
- ChatQnA/docker_compose/intel/hpu/gaudi/compose_guardrails.yaml
- ChatQnA/docker_compose/intel/hpu/gaudi/compose_tgi.yaml
- ChatQnA/docker_compose/intel/hpu/gaudi/compose_without_rerank.yaml

Please verify if the helm charts need to be changed accordingly.

> This issue was created automatically by CI.

yongfengdu:
Don't need to define NUM_CARDS, helm chart already support multiple card running with bigger models.
Refer to https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/agentqna/gaudi-values.yaml (vLLM) and https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/agentqna/gaudi-tgi-values.yaml (tgi)

PR #801, opened 2025-02-12
Support multiple card for LLM inference

yongfengdu:
DeepSeek models require more than one Gaudi cards. Defined NUM_CARDS for user to configure.
Pass NUM_CARDS to container as compose does.
Pass NUM_SHARD to tgi and --tensor-parellel-size for vllm. Fix CI detect helm chart issue.
Remove useless livenessProbe.

## Description

Add multiple cards support to make it easy to support DeepSeek models.

## Issues

#772 

## Type of change

## Dependencies

## Tests

Run on Gaudi environment with this value file:
$cat ds-values.yaml
vllm:
  LLM_MODEL_ID: deepseek-ai/DeepSeek-R1-Distill-Llama-70B
    #LLM_MODEL_ID: deepseek-ai/DeepSeek-R1-Distill-Qwen-32B
  NUM_CARDS: 4
  resources:
    limits:
      habana.ai/gaudi: 4


yongfengdu:
AgentQnA failure:
https://hub.docker.com/r/opea/embedding/tags this image must be wrong. the 1.2 version is 136M, the latest one is more than 3G.
Reported issue to CI team

eero-t (changes requested):
If multiple cards are needed, specifying count of them only in LLM CLI args is not enough, correct number of device resources needs to be requested also in  `resources` section of the deployments.

yongfengdu:
The default is still using NUM_CARD: 1.
See the "Tests" part, which I pasted the example configurations of using 4 devices, including both NUM_CARDS and resources.limit.
AgentQnA is another example of using 4 gaudi cards.
> If model needs multiple devices, specifying count of them only in LLM CLI args is not enough, correct number of devices needs to be requested also in `resources` section of the deployments.

eero-t:
> The default is still using NUM_CARD: 1. See the "Tests" part, which I pasted the example configurations of using 4 devices, including both NUM_CARDS and resources.limit. AgentQnA is another example of using 4 gaudi cards.

Sure, but if you make it configurable, you need to configure the resource request to match it (use that same variable for it).

yongfengdu:
Yeah, I see what's your concern.
The purpose of this PR is to quickly align with above docker compose changes (Pass ENV and leverage that).
To use one configuration option to control all CLI parameters and resource.limit definition will require more work, we can do it later.(Design from user's perspective) but will not include in this PR.
> > The default is still using NUM_CARD: 1. See the "Tests" part, which I pasted the example configurations of using 4 devices, including both NUM_CARDS and resources.limit. AgentQnA is another example of using 4 gaudi cards.
> 
> Sure, but if you make it configurable, you need to configure the resource request to match it (use that same variable for it).

lianhao:
CI is failing due to known issue https://github.com/opea-project/GenAIComps/issues/1295

eero-t:
> To use one configuration option to control all CLI parameters and resource.limit definition will require more work, we can do it later.

@yongfengdu At least move the NUM_CARDS value setting to be just after the resource requests, so that user notices both needing to be updated at the same time.

yongfengdu:
Done
> @yongfengdu At least move the NUM_CARDS value setting to be just after the resource requests, so that user can more easily notice both needing to be updated at the same time.

eero-t (approved):
Thanks, looks fine now!

lianhao (approved):

yongfengdu:
Considering @eero-t  comments and after discussion with @lianhao, to define a NUM_CARDS doesn't help improving user's experience, nor simplifying the helm chart implementation.
So I'd like to close this PR.
Multiple cards support is already supported, refer to https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/agentqna/gaudi-values.yaml (vLLM) and https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/agentqna/gaudi-tgi-values.yaml (tgi)

PR #804, opened 2025-02-13
Fix wrong detection of helm-chart name in CI

yongfengdu:
Filter files that are not helm charts

## Description

Fix the failure of update helm charts
https://github.com/opea-project/GenAIInfra/actions/runs/13257577897/job/37006999398

## Issues

## Type of change

## Dependencies

## Tests

### DIFF
--- a/.github/workflows/pr-chart-e2e.yaml
+++ b/.github/workflows/pr-chart-e2e.yaml
@@ -42,7 +42,7 @@ jobs:
           merged_commit=$(git log -1 --format='%H')
           e2e_charts=$(git diff --name-only ${base_commit} ${merged_commit} | \
           grep "^$CHARTS_DIR/" | \
-          grep -vE 'README.md|CI.md|common|*.sh' | \
+          grep -vE 'valuefiles.yaml|common|*.md|*.sh' | \
           cut -d'/' -f2 | sort -u )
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-chart-e2e.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           merged_commit=$(git log -1 --format='%H')
           e2e_charts=$(git diff --name-only ${base_commit} ${merged_commit} | \
           grep "^$CHARTS_DIR/" | \
-          grep -vE 'README.md|CI.md|common|*.sh' | \
+          grep -vE 'valuefiles.yaml|common|*.md|*.sh' | \
```
mkbhanda, line 45: does not valuefiles also have to be a pattern? we have things like values_gaudi.yaml and just values.yaml etc.

yongfengdu: This valuefiles.yaml is not values.yaml for helm chart, it's a record of *values.yaml files, which are copied to GenAIExamples/GenAIComps, so we can keep them sync between repos.

lianhao (approved):

yongfengdu:
Updated
> Check could be for chart dirs, but that's small detail. Approved.

mkbhanda (approved):
LGTM

PR #811, opened 2025-02-14
LLM TextGen Bedrock Support

jonminkin97:
## Description

Provide support for using Amazon Bedrock with llm-textgen through environment variable configuration.

## Issues

n/a

## Type of change

## Dependencies

n/a

## Tests

Performed `helm install llm-uservice llm-uservice -f llm-uservice/variant_textgen-bedrock-values.yaml` and confirmed that inference is still functional. Performed installation using both AWS access keys and using EKS Pod Identity to provide AWS credentials to container.

### DIFF
--- a/helm-charts/common/llm-uservice/README.md
+++ b/helm-charts/common/llm-uservice/README.md
@@ -60,7 +60,7 @@ Open another terminal and run the following command to verify the service if wor
 # for llm-textgen service
 curl http://localhost:9000/v1/chat/completions \
   -X POST \
-  -d '{"model": "${LLM_MODEL_ID}", "messages": "What is Deep Learning?", "max_tokens":17}' \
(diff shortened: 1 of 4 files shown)


helm-charts/common/llm-uservice/README.md
lianhao, line 1: After line 36, maybe we should give user an example of how to deploy llm-textgen with bedrock? Just like tgi/vllm?

In line 87, how about add BEDROCK as the third candidate value for setting TEXTGEN_BACKEND?

lianhao (approved):

jonminkin97:
@yongfengdu could you please take a look at the proposed changes when you get a chance? Thank you.

yongfengdu (approved):

Issue #815, opened 2025-02-17
Charts' device value files override component security contexts


eero-t:
Several application values files change component `securityContext`, for example setting (container) root FS to be writable:
```
$ git grep -e '^[^#].*securityContext' -e ^[^#].*readOnlyRoot.*false | grep -v -e common/ -e templates
chatqna/gaudi-tgi-values.yaml:  securityContext:
chatqna/gaudi-tgi-values.yaml:    readOnlyRootFilesystem: false
chatqna/gaudi-vllm-values.yaml:  securityContext:
chatqna/gaudi-vllm-values.yaml:    readOnlyRootFilesystem: false
chatqna/guardrails-gaudi-values.yaml:  securityContext:
chatqna/guardrails-gaudi-values.yaml:    readOnlyRootFilesystem: false
searchqna/gaudi-values.yaml:  securityContext:
searchqna/gaudi-values.yaml:    readOnlyRootFilesystem: false
```

Those overrides are for `tei` & `teirerank` components, which seem to have same override for Gaudi, but I think that's wrong.  Either component needs to write root FS or not, it should not depend on which device is used:
```
$ git grep ^[^#].*readOnlyRoot.*false | grep common/
common/agent/values.yaml:  readOnlyRootFilesystem: false
common/chathistory-usvc/values.yaml:  readOnlyRootFilesystem: false
common/data-prep/values.yaml:  readOnlyRootFilesystem: false
common/llm-uservice/values.yaml:  readOnlyRootFilesystem: false
common/lvm-serve/values.yaml:  readOnlyRootFilesystem: false
common/lvm-uservice/values.yaml:  readOnlyRootFilesystem: false
common/mm-embedding/values.yaml:  readOnlyRootFilesystem: false
common/prompt-usvc/values.yaml:  readOnlyRootFilesystem: false
common/retriever-usvc/values.yaml:  readOnlyRootFilesystem: false
common/speecht5/values.yaml:  readOnlyRootFilesystem: false
common/tei/gaudi-values.yaml:  readOnlyRootFilesystem: false        <====
common/teirerank/gaudi-values.yaml:  readOnlyRootFilesystem: false  <====
common/whisper/values.yaml:  readOnlyRootFilesystem: false
```

I'm pretty sure those overrides were made redundant with https://github.com/opea-project/GenAIInfra/pull/613 and https://github.com/opea-project/GenAIInfra/pull/642.

--------
[...]

PR #867, opened 2025-03-11
Add readOnlyRootFileSystem for dataprep and retriever

lianhao:
## Description

Add readOnlyRootFileSystem for dataprep and retriever.

## Issues

Part of #815.

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/common/retriever-usvc/templates/deployment.yaml
+++ b/helm-charts/common/retriever-usvc/templates/deployment.yaml
@@ -82,6 +82,8 @@ spec:
           volumeMounts:
             - mountPath: /tmp
               name: tmp
+            - mountPath: /home/user/nltk_data
+              name: nltk-data
           {{- if .Values.livenessProbe }}
           livenessProbe:
             {{- toYaml .Values.livenessProbe | nindent 12 }}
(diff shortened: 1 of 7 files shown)


helm-charts/common/retriever-usvc/values.yaml
```diff
@@ -69,7 +69,7 @@ podSecurityContext: {}
 
 securityContext:
   # NOTE: many dependent python modules(e.g. numpy, mpl, haystack, etc.) need to write to local disk
```
poussa, line 71: Should the comment be adjusted now that the FS is r/o?

helm-charts/common/retriever-usvc/templates/deployment.yaml
```diff
@@ -82,6 +82,8 @@ spec:
           volumeMounts:
             - mountPath: /tmp
               name: tmp
+            - mountPath: /home/user/nltk_data
```
poussa, line 85: Do we have this directory? Where it is created?

lianhao: nltk_data directory is created by the service itself during the runtime

helm-charts/common/retriever-usvc/values.yaml
lianhao: done

poussa (approved):

yongfengdu (approved):

PR #874, opened 2025-03-14
vllm: Add securityContxt settings

lianhao:
## Description

1. Add securityContext settings for image of opea/vllm and opea/vllm-gaudi.
2. Increase cpu limits settings for tei/teirerank since ingesting pdf files requires more cpu times to avoid request failure due to
embedding/reranking service unresponsiveness.

## Issues

Related to issue #815

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/agentqna/gaudi-values.yaml
+++ b/helm-charts/agentqna/gaudi-values.yaml
@@ -20,6 +20,9 @@ vllm:
   VLLM_SKIP_WARMUP: true
   shmSize: 16Gi
   extraCmdArgs: ["--tensor-parallel-size", "4", "--max-seq_len-to-capture", "16384", "--enable-auto-tool-choice", "--tool-call-parser", "llama3_json"]
+  securityContext:
+    runAsUser: 0
+    runAsNonRoot: false
 
 supervisor:
   llm_endpoint_url: http://{{ .Release.Name }}-vllm
--- a/helm-charts/common/vllm/gaudi-values.yaml
+++ b/helm-charts/common/vllm/gaudi-values.yaml
@@ -12,3 +12,11 @@ extraCmdArgs: ["--tensor-parallel-size","1","--block-size","128","--max-num-seqs
 resources:
   limits:
     habana.ai/gaudi: 1
+
+# NOTE: opea/vllm-gaudi image requires running as root during runtime
+securityContext:
+  runAsUser: 0
+  runAsNonRoot: false
+# NOTE: opea/vllm-gaudi needsd more warm up time
+startupProbe:
+  failureThreshold: 360
--- a/helm-charts/chatqna/cpu-tgi-values.yaml
+++ b/helm-charts/chatqna/cpu-tgi-values.yaml
@@ -65,7 +65,7 @@ teirerank:
   # Potentially suitable values for scaling CPU TEI v1.5 with BAAI/bge-reranker-base model:
   resources:
     limits:
-      cpu: 4
+      cpu: 24
       memory: 30Gi
     requests:
       cpu: 2
@@ -91,7 +91,7 @@ tei:
   # Potentially suitable values for scaling CPU TEI 1.5 with BAAI/bge-base-en-v1.5 model:
   resources:
     limits:
-      cpu: 4
+      cpu: 24
       memory: 4Gi
     requests:
       cpu: 2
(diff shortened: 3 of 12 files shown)


poussa:
We need `fsGroup` as well. NOTE: it is only pod level setting.

lianhao:
> We need `fsGroup` as well. NOTE: it is only pod level setting.

fsGroup is not a requirement in [pod-security-standards](https://kubernetes.io/docs/concepts/security/pod-security-standards). Also we want to leave the fsGroup settings to the end users because it's their responsibility to setup the PVC to be used as a volume, and fsGroup settings is related to that

yongfengdu (approved):

helm-charts/common/vllm/gaudi-values.yaml
```diff
@@ -12,3 +12,11 @@ extraCmdArgs: ["--tensor-parallel-size","1","--block-size","128","--max-num-seqs
...
   limits:
     habana.ai/gaudi: 1
+
+# NOTE: opea/vllm-gaudi image requires running as root during runtime
+securityContext:
+  runAsUser: 0
+  runAsNonRoot: false
+# NOTE: opea/vllm-gaudi needsd more warm up time
+startupProbe:
+  failureThreshold: 360
```
eero-t, line 22: Value in `helm-charts/common/vllm/values.yaml` i.e. for CPU is:
`failureThreshold: 120`

Is vLLM startup supposed to be 3x faster on CPU than on Gaudi?

lianhao: much more fater than 3x on CPU with the same settings.

yongfengdu: Set "VLLM_SKIP_WARMUP: true" if you want a quick startup on Gaudi, otherwise warmup takes a lot of time.

eero-t: > much more faster than 3x on CPU with the same settings.

This is something that definitely needs to fixed for vLLM to be reasonable choice for HPA scaling.

> Set "VLLM_SKIP_WARMUP: true" if you want a quick startup on Gaudi, otherwise warmup takes a lot of time.

Warmup is supposed to optimize model etc for the underlying HW.  How much performance skipping it looses, Does it e.g. make vLLM significantly slower than TGI?

helm-charts/agentqna/gaudi-values.yaml
```diff
@@ -20,6 +20,9 @@ vllm:
   VLLM_SKIP_WARMUP: true
   shmSize: 16Gi
   extraCmdArgs: ["--tensor-parallel-size", "4", "--max-seq_len-to-capture", "16384", "--enable-auto-tool-choice", "--tool-call-parser", "llama3_json"]
+  securityContext:
+    runAsUser: 0
```
eero-t, line 24: Is the `root` user used because deployment does not know what user can access /dev/accel/* device files?

In that case, please see: https://kubernetes.io/blog/2021/11/09/non-root-containers-and-devices/

eero-t (approved):

helm-charts/agentqna/gaudi-values.yaml
lianhao: > Is the `root` user used because deployment does not know what user can access /dev/accel/* device files?
> 
> In that case, please see: https://kubernetes.io/blog/2021/11/09/non-root-containers-and-devices/

Nope. This is vllm-gaudi specific issue. tgi-gaudi runs well with non-root user on the same k8s node. I believe I've already  done the steps mentioned in the above blog

eero-t: > This is vllm-gaudi specific issue. tgi-gaudi runs well with non-root user on the same k8s node.

Could you file a bug against Habana vLLM fork, if there isn't one yet, and add link to it as comment e.g. to `vllm/gaudi-values.yaml`?

lianhao: Bug is filed in upstream, https://github.com/HabanaAI/vllm-fork/issues/929. Will add the comment in following PRs. I'd like to get this merged in first to unblock incoming dependent PRs

Ruoyu-y (approved):
LGTM

PR #882, opened 2025-03-19
vllm: support to run as non-root

lianhao:
## Description

Both latest images of `opea/vllm` and `opea/vllm-gaudi` needs some special settings to run as non-root.

## Issues

`n/a`.

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/common/vllm/templates/configmap.yaml
+++ b/helm-charts/common/vllm/templates/configmap.yaml
@@ -21,6 +21,7 @@ data:
   NUMBA_CACHE_DIR: "/tmp"
   HF_HOME: "/tmp/.cache/huggingface"
   XDG_CONFIG_HOME: "/tmp"
+  TORCHINDUCTOR_CACHE_DIR: "/tmp/pytorchinductor_cache"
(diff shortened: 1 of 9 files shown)


yongfengdu (approved):

helm-charts/common/vllm/templates/configmap.yaml
```diff
@@ -21,6 +21,7 @@ data:
   NUMBA_CACHE_DIR: "/tmp"
   HF_HOME: "/tmp/.cache/huggingface"
   XDG_CONFIG_HOME: "/tmp"
+  TORCHINDUCTOR_CACHE_DIR: "/tmp/pytorchinductor_cache"
```
eero-t, line 24: You could add here a comment such as "PyTorch requires root user if cache files are not owned by user".

eero-t (approved):
Looks good, approved.  Suggested comment could be added before merge.

eero-t:
PR description could refer to earlier PR that added the root user usage, and why that was needed:
> Latest vLLM enabled support for PyTorch inductor which requires root, if cache files are not owned by the user, but that reason was not known before this PR.

poussa (approved):

PR #925, opened 2025-04-01
codegen: limit max output token in test case

lianhao:
## Description

To avoid CI timeout in busy environment, we need to put limit on max_tokens.

## Issues

`n/a`.

## Type of change

## Dependencies

## Tests


yongfengdu (approved):

Ruoyu-y (approved):
LGTM

PR #931, opened 2025-04-02
Upgrade tgi and tei version

yongfengdu:
## Description

Upgrade the TGI and TEI to be consistent with compose deployment

## Issues

#921 #906 

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

List the newly introduced 3rd party dependency if exists.

## Tests

Describe the tests that you ran to verify your changes.

Issue #922, opened 2025-04-01
[Bug]gpt-sovits CD test failed.


yongfengdu:
### Priority

P2-High

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Installation method

- [ ] Pull docker images from hub.docker.com
- [ ] Build docker images from source
- [x] Other

### Deploy method

- [x] Kubernetes Helm Charts
- [ ] Kubernetes GMC
- [ ] Other

### Running nodes

Multiple Nodes

### What's the version?

Internal CI environment

### Description

https://github.com/opea-project/GenAIInfra/actions/runs/14185909961/job/39741086096
gpt-sovits pod crashing due to this message, need to investigate.
Warning  BackOff    4m58s (x18 over 7m2s)  kubelet            Back-off restarting failed container gpt-sovits in pod gpt-sovits01024648-7b56f74fc6-jkdbb_infra-gpt-sovits-9e96f62e(726cdcb1-716e-4218-8aab-c68ade86c36a)
-----------------------------------
#kubectl logs gpt-sovits01024648-7b56f74fc6-jkdbb -n infra-gpt-sovits-9e96f62e
[pod/gpt-sovits010[246](https://github.com/opea-project/GenAIInfra/actions/runs/14185909961/job/39741086096#step:7:247)48-7b56f74fc6-jkdbb/gpt-sovits] Traceback (most recent call last):
[pod/gpt-sovits01024648-7b56f74fc6-jkdbb/gpt-sovits]   File "/home/user/GPT-SoVITS/api.py", line 153, in <module>
[pod/gpt-sovits01024648-7b56f74fc6-jkdbb/gpt-sovits]     import LangSegment
[pod/gpt-sovits01024648-7b56f74fc6-jkdbb/gpt-sovits]   File "/usr/local/lib/python3.10/site-packages/LangSegment/__init__.py", line 1, in <module>
[pod/gpt-sovits01024648-7b56f74fc6-jkdbb/gpt-sovits]     from .LangSegment import LangSegment,getTexts,classify,getCounts,printList,setLangfilters,getLangfilters,setfilters,getfilters
[pod/gpt-sovits01024648-7b56f74fc6-jkdbb/gpt-sovits] ImportError: cannot import name 'setLangfilters' from 'LangSegment.LangSegment' (/usr/local/lib/python3.10/site-packages/LangSegment/LangSegment.py)
-----------------------------------
+ exit 1

### Reproduce steps

[...]

PR #936, opened 2025-04-03
Remove the livenessProbe for tgi/tei/teirerank

yongfengdu:
Per discussion in #431, livenessProbe might cause restart of slow pods.

## Description

Per discussion of #431 and address comments of #931 

## Issues

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/agentqna/gaudi-tgi-values.yaml
+++ b/helm-charts/agentqna/gaudi-tgi-values.yaml
@@ -1,4 +1,4 @@
-# Copyright (C) 2025 Intel Corporation
+# Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 # Accelerate inferencing in heaviest components to improve performance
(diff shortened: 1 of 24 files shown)


yongfengdu:
The case failed is known issue of opea/gpt-sovits image #922

mkbhanda (commented):
Unrelated to the Liveness probe is how the copyright information was updated.
We should not do away with 2024 when the files were originally created. See https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html and https://softwareengineering.stackexchange.com/questions/34189/how-do-you-update-copyright-notices  for some information on the practice. Copying @chensuyue as an FYI in case we have some scripts that are updating copyright.

lianhao:
(audioqna, cpu-multilang-values) CI test fails due to known issue #922

lianhao (approved):

yongfengdu:
Thanks for catching this. Will change them back.
> Unrelated to the Liveness probe is how the copyright information was updated. We should not do away with 2024 when the files were originally created. See https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html and https://softwareengineering.stackexchange.com/questions/34189/how-do-you-update-copyright-notices for some information on the practice. Copying @chensuyue as an FYI in case we have some scripts that are updating copyright.

mkbhanda (approved):
LGTM

Issue #839, opened 2025-02-28
[ci-auto] Add milvus support in chatqna helm chart


CICD-at-OPEA:
Based on this [docker-compose example](https://github.com/opea-project/GenAIExamples/commit/fba0de45d24c5277ee18c4913af711ffb430863c) for ChatQnA, we need to add milvus DB support in chatqna's helm chart, by following [this example PR to add qdrant](https://github.com/opea-project/GenAIInfra/pull/813) in chatqna.

The milvus DB support is already enabled the `data-prep` and `retriever-usvc` component helm charts.

cmellazchy:
OPEA_Team3 is working on this.

lianhao:
Done by #940

PR #940, opened 2025-04-03
Enable milvus for chatqna VectorDB

yongfengdu:
## Description

Enable milvus for chatqna VectorDB.

## Issues

#839 

## Type of change

## Dependencies

## Tests

### DIFF
--- a/helm-charts/chatqna/cpu-milvus-values.yaml
+++ b/helm-charts/chatqna/cpu-milvus-values.yaml
@@ -1,0 +1,14 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+redis-vector-db:
+  enabled: false
+milvus:
+  enabled: true
+
+data-prep:
+  DATAPREP_BACKEND: "MILVUS"
+  COLLECTION_NAME: "rag_milvus"
+retriever-usvc:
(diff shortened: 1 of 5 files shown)


helm-charts/chatqna/cpu-milvus-values.yaml
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+redis-vector-db:
+  enabled: false
+milvus:
+  enabled: true
```
lianhao, line 7: Should we apply something like the [milvus helm chart settings](https://github.com/opea-project/GenAIInfra/blob/main/helm-charts/common/data-prep/milvus-values.yaml), to match it with GenAIExample's [milvus setting](https://github.com/opea-project/GenAIExamples/blob/main/ChatQnA/docker_compose/intel/cpu/xeon/compose_milvus.yaml)?

yongfengdu: The etcd and minio is managed by milvus helm chart, and the configuration is already included in the values.yaml file. So we just need to enable it in the cpu-mivus-values.yaml to overwrite the "enabled" var, if no requirements of further configuraton(PVC etc).

lianhao (approved):
LGTM

PR #973, opened 2025-04-10
Agent UI using different framework

yongfengdu:
Open WebUI based and use 8080 as default port.
No need for nginx reverse proxy.
Skip auth login by setting WEBUI_AUTH.
Set OPENAI_API_BASE_URL as supervisor url to avoid user configuration.

## Description

The summary of the proposed changes as long as the relevant motivation and context.

## Issues

Fixed #960 

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

List the newly introduced 3rd party dependency if exists.

## Tests

Describe the tests that you ran to verify your changes.

PR #981, opened 2025-04-11
chatqna: add VLLM_CPU_KVCACHE_SPACE setting

### LINKED ISSUE #963: [ci-auto] GenAIExample ChatQnA compose.yaml got changed.
## GenAIExample ChatQnA compose.yaml got changed

Below files are changed in [this commit](https://github.com/opea-project/GenAIExamples/commit/7b7728c6c30f5359eaf7e95d5cace048b1159021)

- ChatQnA/docker_compose/intel/cpu/xeon/compose.yaml

Please verify if the helm charts need to be changed accordingly.

> This issue was created automatically by CI.

lianhao:
## Description

Adapt latest chatqna changes in opea-project/GenAIExamples#1762.

## Issues

Fixes #963.

## Type of change

## Dependencies

## Tests


yongfengdu (approved):

lianhao:
need to wait for PR #973 to land in first

helm-charts/chatqna/values.yaml
```diff
@@ -71,6 +71,7 @@ vllm:
   LLM_MODEL_ID: meta-llama/Meta-Llama-3-8B-Instruct
   shmSize: 128Gi
   VLLM_TORCH_PROFILER_DIR: "/tmp/vllm_profile"
+  VLLM_CPU_KVCACHE_SPACE: 40
```
eero-t, line 74: This means 40GB sized KV cache: https://docs.vllm.ai/en/latest/getting_started/installation/cpu.html#related-runtime-environment-variables

=> Memory request should be set accordingly (i.e. 40GB + enough memory for the model itself).

eero-t (changes requested):

helm-charts/chatqna/values.yaml
eero-t: It would be good to group together all related variables; model ID, KV cache size, and requests, prefixed with e.g. "model specific values"  comment, and the other variables being prefixed with "model independent values" comment.

More complex solution would be including model "profiles" into Helm charts as mappings, and selecting suitable values from that mapping using the specified model ID as a key.

(Llong term solution could be switching to an operator that supports model profiles files or CRDs, and handles these automatically based on selected model and other tags, instead of them being managed with Helm variables.)

lianhao: Seems this setting is specific for DeepSeek model on cpu, I'll move the related settings into cpu-values.yaml, and mark it optional.

poussa (approved):

Ruoyu-y (approved):

helm-charts/chatqna/cpu-values.yaml
```diff
@@ -1,5 +1,10 @@
...
 
-image:
-  repository: opea/chatqna
+vllm:
+  LLM_MODEL_ID: meta-llama/Meta-Llama-3-8B-Instruct
+  # Uncomment the following model specific settings for DeepSeek models
+  #VLLM_CPU_KVCACHE_SPACE: 40
+  #resources:
+  #  requests:
+  #    memory: 60Gi # 40G for KV cache, and 20G for DeepSeek-R1-Distill-Qwen-7B, need to adjust it for other models
```
mkbhanda, line 10: @lianhao - Would 7B not require 28 GB, so adding 30 G better?

mkbhanda: Thank you @lianhao for clarification that you tested the requirement before proposing and it turns out the default for LLMs is to use BF16. which mean parameters * 2 G.

mkbhanda (approved):
Thank you @lianhao !

helm-charts/chatqna/cpu-values.yaml
lianhao: vllm by default will use bf16/fp16. So 20G should be enough for 7B I guess

PR #1001, opened 2025-04-16
Update Helm install and instances for ChatQnA terraform

ezelanza:
## Description

README Updated to solve UI error and update on intel aws instances

## Issues

UI wasn't working due to error on the load balancer configuration

## Type of change

- Update instance to newest intel instances i7ie
- Improvement on HELM install to set nginx as LoadBalancer to fix UI error

- [ x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/cloud-service-provider/aws/eks/terraform/README.md
+++ b/cloud-service-provider/aws/eks/terraform/README.md
@@ -11,7 +11,7 @@ This guide shows how to deploy OPEA applications on Amazon Web Service (AWS) Ela
 
 The setup uses Terraform to create EKS cluster with the following properties:
 
-- 1-node EKS cluster with 50 GB disk and `m7i.x8large` SPOT instance (16 vCPU and 32 GB memory)
+- 1-node EKS cluster with 50 GB disk and `i7ei.x12large` (or `i7ei.x6large`) SPOT instance (16 vCPU and 32 GB memory)
 - Cluster autoscaling up to 10 nodes
 - Storage Class (SC) `efs-sc` and Persistent Volume Claim (PVC) `model-volume` for storing the model data
 - `LoadBalancer` address type for the service for external consumption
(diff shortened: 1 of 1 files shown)


cloud-service-provider/aws/eks/terraform/README.md
```diff
@@ -11,7 +11,7 @@ This guide shows how to deploy OPEA applications on Amazon Web Service (AWS) Ela
 
 The setup uses Terraform to create EKS cluster with the following properties:
 
-- 1-node EKS cluster with 50 GB disk and `m7i.x8large` SPOT instance (16 vCPU and 32 GB memory)
+- 1-node EKS cluster with 50 GB disk and `i7ei.x12large` (or `i7ei.x6large`) SPOT instance (16 vCPU and 32 GB memory)
```
mkbhanda, line 14: Is this an "or"? or no longer m7i.x8large work? And why is that? larger model ?  Newest instance type makes sense but we really have a bunch of options .. should our documentation list the options?

ezelanza: Agree that we should have options based on the models to validate which instance would work best with each model.

In this case , I created the PR mainly to fix the helm error and since the guide referred to a 4th gen of Xeon instances I also updated to have the "Same" instance but in a newer gen (i7ei are 5th gen which comes with better performance). I added "or" because m7 has 8xl but i7ei has only 12xl and 6xl options, this is why I leave it as "or". Any case, the importance of this PR is to fix the helm 

https://aws.amazon.com/ec2/instance-types/i7ie/

poussa (approved):

lianhao (approved):

CICD-at-OPEA:
This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days.

CICD-at-OPEA:
This PR was closed because it has been stalled for 7 days with no activity.

PR #1054, opened 2025-05-16
Add KubeAI monitoring support + vLLM dashboard

eero-t:
## Description

Add KubeAI monitoring support + vLLM dashboard.

Monitoring can be added either by using the helper script, or by calling Helm directly with the new metrics.

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

Manually tested earlier.  TODO: test this version too.


kubeai/README.md
```diff
@@ -165,3 +166,31 @@ curl "http://localhost:8000/openai/v1/chat/completions" \
...
+
+# Observability
+
+With [Prometheus](../helm-charts/monitoring.md) running, install script can enable monitoring of the vLLM inference engine instances.
+
+Script requires Prometheus Helm chart release name for that, e.g.:
+
+```
+release=prometheus-stack
+./install.sh $release
```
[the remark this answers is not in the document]
eero-t: NAK. It's consistent with earlier example.

poussa (approved):

eero-t:
@marquiz, @mkbhanda OK to merge?

kubeai/install.sh
```diff
@@ -5,18 +5,41 @@
...
+(configMap) is installed for Grafana.
+
+ERROR: $1!
+EOF
     exit 1
+}
+
+metrics=""
+for arg in "$@"; do
+	if [ -f "$arg" ]; then
```
marquiz, line 28: nit: the arg parsing looks somewhat shaky'n'shady but I guess that's ok for this kind of hack/helper script

marquiz (approved):
Thanks @eero-t. One nit but I can live with that :smile: I think we can merge this

kubeai/install.sh
eero-t: I would think it very unlikely that user's Prometheus release is named exactly the same as some file in the `kubeai/` directory...

PS. I'm wondering about the benefit of the script, when more things are needed to configure. I think it would be clearer if user would just invoke Helm directly (with command copy-pasted from README), in this case with additional `-f monitoring.yaml` argument.

PR #1089, opened 2025-06-03
Add 4 llm model CRs for kubeai on gaudi platform

PeterYang12:
## Description

Add 4 llm model CRs for kubeai on gaudi platform

## Issues

N/A

## Dependencies

N/A


eero-t:
@poussa After profiles have been merged for all the models we want to support in 1.4 release, I think a model/profile matrix table like this should also be added to the README:
https://github.com/opea-project/Enterprise-Inference/blob/main/docs/supported-models.md

IMHO it should list also how many devices the model (profile) needs.  Maybe also what are the currently specified min/max replicas, so that one knows how many fit into given cluster, if they're deployed as-is.

PR #993, opened 2025-04-14
Add support for external OpenAI-compatible LLM endpoints across Helm charts (chatqna, codegen, docsum)

devpramod:
## Description

This PR implements support for using external OpenAI-compatible LLM endpoints with ChatQnA, CodeGen, and DocSum Helm charts while maintaining backward compatibility with existing configurations.

Motivation
Users need the flexibility to connect to external LLM services (like OpenAI or self-hosted OpenAI-compatible endpoints) instead of always relying on the included LLM components. This adds versatility to our Helm charts without disrupting existing functionality.

Key Changes
For each Helm chart (ChatQnA, CodeGen, DocSum):

1. Added `external-llm-values.yaml` configuration file:
- Provides settings for external LLM endpoints (LLM_SERVER_HOST_IP, LLM_MODEL, OPENAI_API_KEY)
- Disables internal LLM services (tgi, vllm, llm-uservice) when using external endpoints

2. Created `templates/external-llm-configmap.yaml`:
- Manages environment variables required for external LLM integration
- Conditionally created only when external LLM is enabled

3. Updated `templates/deployment.yaml`:
- Added conditional logic to use ConfigMap for environment variables when external LLM is enabled
- Maintained all existing functionality and conditional paths for internal LLM services

3. Updated `Chart.yaml`:
- Added proper conditions for all dependencies to make them optional
- Enables selective component deployment based on configuration values

4. Updated `README.md`:
- Added example commands showing how to use external LLM endpoints
- Maintained consistent placeholder values across all services

## Issues

n/a

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

'n/a'

## Tests

Tested backward compatibility to ensure existing helm charts with default values.yaml is working

PR #1140, opened 2025-07-01
Extend CI scripts to test for external LLM endpoints

devpramod:
## Description

This PR extends the CI scripts to support testing with external LLM endpoints. The changes add configuration for external LLM services in the Helm E2E workflow, allowing charts with external-llm-values.yaml files to be tested against external LLM providers. The implementation includes environment variable mapping for different chart types (ChatQnA, DocSum, CodeGen etc) and supports configurable LLM server hosts, ports, models, and API keys.

## Issues

n/a

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

No new 3rd party dependencies introduced.

## Tests

Describe the tests that you ran to verify your changes.
The changes extend existing Helm E2E tests to support external LLM endpoint testing. Charts with external-llm-values.yaml files will now be tested with the configured external LLM services using the appropriate environment variables and Helm chart arguments.

PR #1153, opened 2025-07-04
Fix READMEs to match (external LLM) values file names

eero-t:
## Description

Fix #993 values file names to match README.

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

CI.


eero-t:
> I think we need to fix README, the `variant_` prefix here is meant to temporarily skip CI

Was the merged https://github.com/opea-project/GenAIInfra/pull/1140 supposed to fix CI for external endpoints?

lianhao:
> > I think we need to fix README, the `variant_` prefix here is meant to temporarily skip CI
> 
> Btw. was the merged #1140 supposed to fix CI for external endpoints?

Nope. We also need to Suyue's help to setup the corresponding secrets to be used by #1140

Issue #1163, opened 2025-07-15
[Bug] keep restarting failed container model-downloader in pod chatqna-teirerank, chatqna-tei for ChatQnA


louie-tsai:
### Priority

P2-High

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Installation method

- [x] Pull docker images from hub.docker.com
- [ ] Build docker images from source
- [ ] Other
- [ ] N/A

### Deploy method

- [x] Kubernetes Helm Charts
- [ ] Kubernetes GMC
- [ ] Other
- [ ] N/A

### Running nodes

Single Node

### What's the version?

the latest main branch

### Description

trying ChatQnA on customer K8s cluster, but keep restarting failed container model-downloader in pod chatqna-teirerank, chatqna-tei for ChatQnA.

following the instruction for Xeon, but facing model-download issue.
https://github.com/opea-project/GenAIInfra/tree/main/helm-charts/chatqna#installing-the-chart

<img width="1456" height="557" alt="Image" src="https://github.com/user-attachments/assets/a9ded80a-7eeb-4896-a9e9-cbbab87469ee" />

### Reproduce steps

following instructions for ChatQnA with vLLM on Xeon
https://github.com/opea-project/GenAIInfra/tree/main/helm-charts/chatqna#installing-the-chart

### Raw log

```shell

```

louie-tsai:
thanks.  saw the issue from the model downloader container.

<img width="936" height="297" alt="Image" src="https://github.com/user-attachments/assets/8297d7c3-0c73-42b9-9438-99d67160d172" />

PR #1164, opened 2025-07-16
README.md: set HF_TOKEN also in remote endpoint case

louie-tsai:
## Description

Remote Endpoint case miss HF Token for TEI and TEI Reranking.

## Issues

[#1163 ]

## Type of change

## Dependencies
NA

## Tests

Test on customer cluster


eero-t (approved):
Approved.  I just tested this, and bumped on it also myself.

Btw. Another PR would be needed for adding external support also for OpenAI API embedding inferencing endpoint, not just for LLM inferencing endpoint.

(AFAIK reranking inferencing does not have yet OpenAI API, so that would need to use some other API.)

lianhao (approved):

